PR #491, opened 2020-07-25
Fix power becoming 0 when killing another player

Pasarus:


src/main/java/factionsystem/Commands/CreateCommand.java
```diff
@@ -48,7 +48,7 @@ public boolean createFaction(CommandSender sender, String[] args) {
                 if (!factionExists) {
 
                     // actual faction creation
-                    Faction temp = new Faction(name, player.getUniqueId(), main.getConfig().getInt("maxPowerLevel"));
+                    Faction temp = new Faction(name, player.getUniqueId(), main.getConfig().getInt("maxPowerLevel"), main);
```
dmccoystephenson, line 51: Now that we're passing main in we won't need to pass the max power level in

PR #492, opened 2020-07-25
Fix faction home not being invalid on claim

Pasarus:
Needs proper testing


dmccoystephenson:
Tested this with a player on my server and it works perfectly.

Issue #565, opened 2020-08-24
Add a method to delete old config options in the config.yml.


dmccoystephenson:
When a config option is renamed, the old name will need to be deleted using this method.

Rename hourlyPowerIncreaseAmount to powerIncreaseAmount.

Rename maxPowerLevel to initialMaxPowerLevel.

Rename officerLimit to playersNeededForNewOfficer.

dmccoystephenson:
Attempted to do this in my delete-old-config-options branch. The renaming of the config options seems to have gone fine, but there appears to be a problem with this code, as no old config options are getting deleted.

```
    private void deleteOldConfigOptionsIfPresent() {
        for (String option : oldConfigOptions) {
            deleteConfigOption(option);
        }
    }

    // Credit: https://stackoverflow.com/questions/1377279/find-a-line-in-a-file-and-remove-it
    private void deleteConfigOption(String option) {
        try {
            // iterate through lines in config.yml
            File inputFile = new File("/plugins/MedievalFactions/config.yml");
            File tempFile = new File("/plugins/MedievalFactions/temp-config.yml");

            BufferedReader reader = new BufferedReader(new FileReader(inputFile));
            BufferedWriter writer = new BufferedWriter(new FileWriter(tempFile));

            String lineToRemove = option;
            String currentLine;

            while((currentLine = reader.readLine()) != null) {
                // trim newline when comparing with lineToRemove
                String trimmedLine = currentLine.trim();
                if(trimmedLine.contains(lineToRemove)) continue;
                writer.write(currentLine + System.getProperty("line.separator"));
            }
            writer.close();
            reader.close();
            tempFile.renameTo(inputFile);
        }
        catch(Exception e) {
            System.out.println("Something went wrong when deleting a config option.");
        }
    }
```

dmccoystephenson:
Changed the code, still doesn't work.

```
    private void deleteOldConfigOptionsIfPresent() {
        for (String option : oldConfigOptions) {
            deleteConfigOption(option);
        }
    }

    private void deleteConfigOption(String option) {
        File configFile = new File("./plugins/MedievalFactions/config.yml");
        File tempFile = new File("./plugins/MedievalFactions/temp-config.yml");
        System.out.println("Attempting to delete old config option: " + option);
        try {
            Scanner scanner = new Scanner(configFile);
            FileWriter writer = new FileWriter(tempFile);
            String currentLine;

[...]

PR #629, opened 2020-08-31
Issue #565 - Delete Old Config Options + Config Subsystem Addition

dmccoystephenson:
Everything is set up and ready to go... except for the method that does the actual deleting. I have set config-related methods into a Config Subsystem that also has a list of old config options. The console tells me it works but the file manager shows me that it lies.

I don't know why this doesn't work. If you guys want to try to run it and figure it out that would be awesome.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -6,7 +6,7 @@
 
     <groupId>KingdomProgrammers</groupId>
     <artifactId>Medieval-Factions</artifactId>
-    <version>v3.4.3</version>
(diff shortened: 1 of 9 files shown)


dmccoystephenson:
I asked the SpigotMC Support Discord Server, hopefully that leads to something.
https://discordapp.com/channels/690411863766466590/741875863271899136/749892892096725124

dmccoystephenson:
Setting a config option to null deletes it... just for future reference...

PR #637, opened 2020-09-01
Issue 583

philbgarner:


philbgarner:
This PR includes all of the changes in each of the individual branches, only need to apply this one to get all of the changes from each branch.

pom.xml
```diff
@@ -67,7 +67,7 @@
         <dependency>
             <groupId>org.spigotmc</groupId>
             <artifactId>spigot-api</artifactId>
-            <version>1.14.4-R0.1-SNAPSHOT</version>
+            <version>1.16.1-R0.1-SNAPSHOT</version>
```
dmccoystephenson, line 70: Why was this changed to 1.16.1? It is 1.14.1 so that the plugin is supported for 1.14+

It occurs to me this could have been to include some blocks added in 1.16.1. An if check needs to be added in this case for these new blocks to check the version of minecraft being run I think. That way we don't end up with console errors for people running 1.14 or 1.15 servers when they try to lock blocks.

I believe the pom.xml version can remain 1.16.1 as long as these if checks are added.

PR #662, opened 2020-09-12
Fix to remove == comparison on strings.

philbgarner:
This fixes the issue, it was caused by == comparisons to the getWorld() string instead of .equalsIgnoreCase().

Sorry about that, c# is creeping in.


dmccoystephenson (approved):
Sounds good! Thanks for doing this!

PR #687, opened 2020-09-13
made power decreases only affect offline players by adding a ! to an if statement

dmccoystephenson:
Decided to take a quick break from HW and take a look at the power decrease code. Correct me if I'm wrong, but I'm pretty sure that the way it is programmed right now, power decreases will only affect online players. I added a ! to negate the isOnline check.

@philbgarner If you could review this PR whenever you have the chance (doesn't have to be today) that would be awsome. Since you programmed the original method, I want to verify that this merge commit will result in the behavior you intended.

### DIFF
--- a/src/main/java/factionsystem/Subsystems/UtilitySubsystem.java
+++ b/src/main/java/factionsystem/Subsystems/UtilitySubsystem.java
@@ -385,7 +385,7 @@ public class UtilitySubsystem {
     				{
     					isOnline = player.isOnline();
     				}
-    				if (isOnline && main.getConfig().getBoolean("powerDecreases")
+    				if (!isOnline && main.getConfig().getBoolean("powerDecreases")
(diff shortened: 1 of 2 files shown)


dmccoystephenson:
Just realized my branch name is inaccurate. I meant to call the branch "making-power-decreases-only-affect-**offline**-players', so sorry about that.

src/main/java/factionsystem/Subsystems/UtilitySubsystem.java
```diff
@@ -385,7 +385,7 @@ public void run() {
     				{
     					isOnline = player.isOnline();
     				}
-    				if (isOnline && main.getConfig().getBoolean("powerDecreases")
```
dmccoystephenson, line 388: This if statement checked to see if a player was online as one of the conditions for decreasing their power, meaning power decreases don't affect offline players like they should. I negated the isOnline condition.

philbgarner:
Looks great, thanks Dan.

PR #708, opened 2020-09-19
Gates

philbgarner:
Gates are now in a workable state, saves and loads with /mf force save/load, most features are in place.  Some glitches I'm still tweaking, however I wanted to integrate this with all of the newer changes because I'm almost done with the tasks in the project epic.

I pulled a fresh fork a few hours ago from the master branch and manually added the files and references before submitting this pull request so this should be the minimal set of changes that do not override anything new you've added.

I put a build of this up on the test server if you want to play around with it before approving.

### DIFF
--- a/src/main/java/factionsystem/Commands/GateCommand.java
+++ b/src/main/java/factionsystem/Commands/GateCommand.java
@@ -1,0 +1,86 @@
+package factionsystem.Commands;
+
+import static factionsystem.Subsystems.UtilitySubsystem.createStringFromFirstArgOnwards;
+
+import java.util.Set;
+
+import org.bukkit.ChatColor;
+import org.bukkit.Material;
+import org.bukkit.command.CommandSender;
+import org.bukkit.entity.Player;
+
+import factionsystem.Main;
+import factionsystem.Objects.Faction;
+import factionsystem.Objects.Gate;
+import factionsystem.Subsystems.UtilitySubsystem;
+
+public class GateCommand {
+
+	Main main = null;
+	
+	public GateCommand(Main plugin)
+	{
+		main = plugin;
+	}
+	
+	public void handleGate(CommandSender sender, String[] args)
+	{
+		if (sender instanceof Player)
+		{
+			Player player = (Player) sender;
+			
+			if (player.hasPermission("mf.gate") || player.hasPermission("mf.default"))
+			{
+				if (args.length > 1)
+				{
+					if (args[1].equalsIgnoreCase("cancel"))
+					{
+						if (main.creatingGatePlayers.containsKey(player.getUniqueId()))
+						{
+							main.creatingGatePlayers.remove(player.getUniqueId());
+							player.sendMessage(ChatColor.RED + "Creating gate cancelled!");
+							return;
+						}
+					}
+					if (args[1].equalsIgnoreCase("list"))
+					{
+						Faction faction = UtilitySubsystem.getPlayersFaction(player.getUniqueId(), main.factions);
+						if (faction != null)
+						{
+							player.sendMessage(ChatColor.AQUA + "Faction Gates");
+							for (Gate gate : faction.getGates())
+							{
+								player.sendMessage(ChatColor.AQUA + String.format("%s: %s", gate.getName(), gate.coordsToString()));
+							}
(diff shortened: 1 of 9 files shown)


src/main/java/factionsystem/Objects/Gate.java
```diff
@@ -0,0 +1,717 @@
...
+    					b.setType(material);
+    				};
+					if (b != null)
+						getWorld().playSound(b.getLocation(), soundEffect, 0.1f, 0.1f);
+				}		
+			}
+		}
+		else
+		{
+			// TODO: Bridge code iterates over x/z
```
dmccoystephenson, line 693: Does this need to be completed before merging?

src/main/java/factionsystem/Subsystems/CommandSubsystem.java
```diff
@@ -613,6 +613,13 @@ public boolean interpretCommand(CommandSender sender, String label, String[] arg
                     return true;
                 }
 
+                // gate management commands
+                if (args[0].equalsIgnoreCase("gate") || args[0].equalsIgnoreCase("gt")) {
+                	GateCommand command = new GateCommand(main);
```
dmccoystephenson, line 618: The details of the gate command needs to be added to the output of the the /mf help command.

dmccoystephenson (changes requested):

src/main/java/factionsystem/Objects/Gate.java
philbgarner: It only means there is no horizontal gate (IE: bridge) support right now, but I figured since the gates were the core requirement this would be more of a nice-to-have.

src/main/java/factionsystem/Commands/GateCommand.java
```diff
@@ -0,0 +1,86 @@
...
+									gate.setName(name);
+								}
+								else
+								{
+									player.sendMessage(ChatColor.AQUA + String.format("%s Gate.", gate.getName()));
+									return;
+								}
+							}
+						}
+					}
```
dmccoystephenson, line 76: If a player uses the /f gate command incorrectly, they need to be told the possible usages of the command.

dmccoystephenson (changes requested):

src/main/java/factionsystem/Objects/Gate.java
dmccoystephenson: I gotchya, that can totally be added later

dmccoystephenson:
I'm not understanding how to create gate. Is it with the /f gate name command?

src/main/java/factionsystem/Commands/GateCommand.java
```diff
@@ -0,0 +1,86 @@
...
+							player.sendMessage(ChatColor.RED + "Creating gate cancelled!");
+							return;
+						}
+					}
+					if (args[1].equalsIgnoreCase("list"))
+					{
+						Faction faction = UtilitySubsystem.getPlayersFaction(player.getUniqueId(), main.factions);
+						if (faction != null)
+						{
+							player.sendMessage(ChatColor.AQUA + "Faction Gates");
```
dmccoystephenson, line 50: If there are no gates in the faction, the player should be told so and also how to create a new gate.

src/main/java/factionsystem/Commands/GateCommand.java
```diff
@@ -0,0 +1,86 @@
...
+						{
+							main.creatingGatePlayers.remove(player.getUniqueId());
+							player.sendMessage(ChatColor.RED + "Creating gate cancelled!");
+							return;
+						}
+					}
+					if (args[1].equalsIgnoreCase("list"))
+					{
+						Faction faction = UtilitySubsystem.getPlayersFaction(player.getUniqueId(), main.factions);
+						if (faction != null)
```
dmccoystephenson, line 48: If the faction is null, the player should be told they need to be in a faction to use the command.

dmccoystephenson (approved):
Sorry for all the requests! This is some awesome work @philbgarner!

dmccoystephenson:
When I unclaim land that a gate is on, the gate remains in the list of gates when I type /f gate list

PR #721, opened 2020-09-21
Addressed issue #718: gate lever cannot be switched if the gate status

philbgarner:
is not "ready".


dmccoystephenson (approved):
Great job!

PR #752, opened 2020-10-03
Bug fixes: losing player head now inserts into winner's inventory or …

philbgarner:
…drops at their feet.  Time limit set to double so math doesn't become an integer and the timer boss bar never decrements. This should fix issue 747.


src/main/java/factionsystem/EventHandlers/EntityDamageByEntityEventHandler.java
```diff
@@ -23,6 +23,7 @@ public void handle(EntityDamageByEntityEvent event) {
    	
         // this method disallows PVP between members of the same faction and between factions who are not at war
         // PVP is allowed between factionless players, players who belong to a faction and the factionless, and players whose factions are at war.
+        System.out.println("EntityDamageByIntity" + event.toString());
```
dmccoystephenson, line 26: Does this need to remain?

dmccoystephenson (changes requested):

dmccoystephenson:
Thanks for doing this!

PR #776, opened 2020-12-08
Refactoring

dmccoystephenson:


dmccoystephenson:
Deployed this on the test server and it seems to work.

PR #781, opened 2020-12-12
Persistent Data Class

dmccoystephenson:


dmccoystephenson:
deployed this on the test server and didn't immediately run into errors so that's good

PR #788, opened 2020-12-14
Method Migration, UUIDChecker, StringBuilder, Messenger, etc...

dmccoystephenson:
No more Utilities class!


dmccoystephenson:
I deployed this on the test server and tested out a few commands. Nothing seems immediately broken.

PR #812, opened 2020-12-22
Tweaking bypass break door lock mechanic

dmccoystephenson:


dmccoystephenson:
Tested this and it works!

PR #814, opened 2020-12-27
Dynmap Integration (thank you Caibinus!)

dmccoystephenson:


dmccoystephenson (commented):
Several thoughts I've had:
- Dynmap should be updated when the server starts up, and on a scheduled basis, rather than whenever land gets claimed or unclaimed.
- We will need to test whether putting the depend statement in the plugin.yml makes the plugin unable to run without that plugin.
- We can add colors later

PR #927, opened 2021-01-26
rewrote vassalization command and added check to ensure no vassalization loops will occur

### LINKED ISSUE #925: Disallow the vassalization of a faction if it will result in a vassalization loop.
A vassalization loop looks like this:

Faction A vassalizes Faction B

Faction B vassalizes Faction C

Faction C vassalizes Faction A

dmccoystephenson:
closes #925


dmccoystephenson:

PR #1029, opened 2021-05-28
added an event handler method to InteractionHandler.java to attempt to prevent item frame rotation when the rotating player shouldn't have interaction access

### LINKED ISSUE #1008: Fix players being able to rotate items in item frames inside claims that they shouldn't have interaction access to.

dmccoystephenson:
closes #1008 


dmccoystephenson:

PR #1030, opened 2021-05-28
cleaned up event handler method for the BlockBreakEvent event

dmccoystephenson:


dmccoystephenson:

PR #1032, opened 2021-05-28
There should now be a singular method for getting claimed chunks with only one parameter, rather than 3-4.

dmccoystephenson:
This should simplify parts of the code where a claimed chunk needs to be grabbed.


dmccoystephenson (commented):
I just reviewed my changes and it doesn't seem like anything should be broken. I'd still like to test the mechanics relating to claimed chunks before merging, though,

PR #1074, opened 2021-06-04
finishDuel() method now has calls to clear active potion effects on participating players

### LINKED ISSUE #992: Fix dueling glitching out if the losing player dies from fire damage.
> "The dueling glitches out if the person dies from fire damage. They die instead of their health resetting, their head does not drop, and the duel continues. In my server there is a plugin that lets people teleport back to where they died so people have been going back and fighting again over and over." - Aedionic

dmccoystephenson:
Caibinus was responsible for this fix. This still needs to be tested before merging.

closes #992


dmccoystephenson:
Just tested this and it seems to work! Merging!

PR #1078, opened 2021-06-05
Expanded the FactionFlags class to include non-boolean values.

### LINKED ISSUE #1075: Expand the FactionFlags class to include non-boolean values.
I believe the FactionFlags class could be expanded to include non-boolean values with the introduction of more hashmaps for more types.

dmccoystephenson:
Whether flags still persist after a restart will need to be tested before merging.

closes #1075


dmccoystephenson:
An unfortunate consequence of these changes is that if any servers are using the most recent beta version, then their flags will be overwritten by the default values. Still, it seems only I use my beta versions so it should have a minimal impact overall.

dmccoystephenson:
I just tested that flags still persist, and they do! Merging!

PR #1090, opened 2021-06-06
fixing "null" appearing in duel acceptance messages

### LINKED ISSUE #1077: Fix "null" appearing in alerts that get sent when a player accepts another player's duel.

dmccoystephenson:
closes #1077


dmccoystephenson:
I just tested this and the fix worked!

PR #1105, opened 2021-06-17
Dynmap Integration Realms Layer

### LINKED ISSUE #1106: Realms Layer
Add a "Realms" layer to the dynmap, allowing users to view top-liege-level information.

dmccoystephenson:
Caibinus is responsible for this new feature.

Note: These changes still need to be tested.

closes #1106


dmccoystephenson:
I have just tested these changes and it seems there may be a problem. When there is only one faction that has claimed land, the claims on the dynmap will occasionally stop showing. When there are two factions that have claimed land, only one of those factions' claims will show at any one time. 

It seems to cycle between displaying them, for some reason. If one faction has vassalized the other, there seems to be a darker border but the cycling still continues, only slightly differently. This ended up happening:
image

philbgarner:
That is really really weird, might be hard to diagnose this one.

dmccoystephenson:
It looks like it's working to me! Merging!

Issue #971, opened 2021-03-27
Special characters in the translated text for the plugin turn into incomprehensible symbols in-game.


Horuse:
![image](https://user-images.githubusercontent.com/39675195/112728102-37a8a980-8f2e-11eb-8c73-01512572d1ec.png)
in the language file itself everything is fine
![image](https://user-images.githubusercontent.com/39675195/112728148-6de62900-8f2e-11eb-91c2-b261dacc6bf1.png)

dmccoystephenson:
Yeah this is an unfortunate one. When programming the Localization feature, I tested it on a server hosted on ServerMiner and had no problems. When I deployed the finished version to my community server hosted on PebbleHost, I ran into this problem. I'm not sure what causes it, unfortunately.

Can you let me know what server provider you're using? @Horuse

Horuse:
I managed to find how to solve the problem
![image](https://user-images.githubusercontent.com/39675195/112750256-48116080-8fd0-11eb-8235-8094f869e69c.png)
need to add this line to your start.bat

PR #1113, opened 2021-07-04
Updated LocaleManager to support UTF-8 Explicitly. (Issue: #971)

CallumJohnson:
Issue: #971

By default, systems such as Windows and some distros of Linux encode their files using UTF-8. I believe from my past in development, some server providers such as PebbleHost do not by default encode with UTF-8. This is backed up by the comment https://github.com/dmccoystephenson/Medieval-Factions/issues/971#issuecomment-808882355 which simply stated that running the project in the UTF-8 System Encoding resolved the issue.

I have therefore updated the LocaleManager to explicitly write/read using UTF-8.

Note: *I do not have access to a PebbleHost provider, but I have tested this locally using Linux & Windows with success.*


dmccoystephenson (approved):
The code looks good and it works! Merging!

PR #1119, opened 2021-07-10
Vassals can now be called into wars

### LINKED ISSUE #793: Make lieges able to call vassals into wars.
The Invoke Command could be extended to include this functionality.

dmccoystephenson:
closes #793


dmccoystephenson:
I haven't tested this but I can't imagine that it won't work.

PR #1160, opened 2021-09-10
Added limit for the length of the Faction Name.

CallumJohnson:
Issues Fixed: #1134 & #1135
Todo for Dan: Locale message which states that the name is too long.


dmccoystephenson:
Thank you very much for your contribution! I removed the associated epic since there are tasks yet to be completed, just so you know. I appreciate you putting the time in!

Issue #1155, opened 2021-09-05
Fix ClassCastException error printing in the console.


dmccoystephenson:
[00:07:36] [Server thread/WARN]: [MedievalFactions] Task #8297972 for MedievalFactions v4.2.1 generated an exception
java.lang.ClassCastException: org.bukkit.craftbukkit.v1_16_R3.block.CraftBlockState cannot be cast to org.bukkit.block.Chest
	at dansplugins.factionsystem.eventhandlers.InteractionHandler$1.run(InteractionHandler.java:130) ~[?:?]
	at org.bukkit.craftbukkit.v1_16_R3.scheduler.CraftTask.run(CraftTask.java:81) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at org.bukkit.craftbukkit.v1_16_R3.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:400) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.b(MinecraftServer.java:1060) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.DedicatedServer.b(DedicatedServer.java:355) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.a(MinecraftServer.java:1008) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.w(MinecraftServer.java:847) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.lambda$0(MinecraftServer.java:164) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_292]

PR #1162, opened 2021-09-11
Reinforced Chest-Handling for LockedBlock mechanisms.

CallumJohnson:
Fixes: #1155

2 Seconds had passed since the previous check of BlockType on the Chest, giving a player enough time to mine/replace the Chest itself.

Also provided a more in-depth check to determine if a Block is in-fact a Chest by instance checking its BlockState.


dmccoystephenson:
Thank you very much for fixing that issue! The code looks good so i’m gonna go ahead and merge this.

Issue #1109, opened 2021-06-17
Flag Validation: dynmapTerritoryColor


dmccoystephenson:
We should inform the player whether or not their inputted color is valid.

Issue #1108, opened 2021-06-17
Flag Extension: Color Names for "dynmapTerritoryColor"


dmccoystephenson:
> Since the dynmapTerritoryColor uses CSS hex code notation, we can improve usability by adding the list of standard minecraft colour names to the dynmap integrator and if you pass that as a flag it swaps it out for the hex code before rendering the polygons. - Caibinus

PR #1187, opened 2021-09-20
Added check to determine if the given color is valid and also added DefaultColors for usage of string names instead of hexadecimal codes.

CallumJohnson:
Linked issues: #1109 & #1108

The DefaultColor enumeration contains a list of Default Minecraft colors and respective color codes which are gathered from an external site ([HTML Colors](https://htmlcolors.com/color-names)). I couldn't find official values, but these can be switched out very easily if you manage to find them.

I have also added a verification check alongside this new enumeration to determine if the given string conforms to the standard of hexadecimal color codes ([Hex Triplet](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet)). Supported values encapsulate the range of hexadecimal color codes supported by both the Web and Minecraft; ranging from `#000` to `#FFF` (supporting F irrespective of case).


dmccoystephenson:
First of all, these changes look great as always. I admire your programming skills and very much appreciate you putting them to use improving this project.

Correct me if I'm wrong, but it seems that only the requirements of #1109 have been met with this pull request. #1108 would allow players to type /mf flags set 'dynmapTerritoryColor' 'blue' and have that be translated to a hex code. As far as I was able to tell when testing, this wasn't possible.

@CallumJohnson 
Could you verify that this functionality is possible?

CallumJohnson:
'Blue' parsed to '#1569C7' or 'R: 21, G: 105 and B: 199'
image

Obviously... the chat message needs to be modified and replaced with a Locale message, as well as a new locale message for when the hexadecimal format / code isn't recognised.

dmccoystephenson:
I must have not tested your code correctly. Thank you for verifying the functionality. Looks great to me!

CallumJohnson:
Test the new Color Conversion class, setting a Bukkit-provided Color definition 'teal'.
Image
As you can see, the color is parsed using Bukkit, then translated into `#008080` ( [Hex Data](https://www.color-hex.com/color/008080) ), or 'R: 0, G: 128, B: 128'.

dmccoystephenson:
Sounds great!

Issue #1286, opened 2022-01-28
Fix factionless players being able to access any locked block if they type /mf lock before interacting with it.


dmccoystephenson:
A player encountered this error on Kingdom: Unleashed (before the v4.5 update)

Mr-Deej:
can be closed @dmccoystephenson

dmccoystephenson:
Thanks! @Mr-Deej

PR #1298, opened 2022-01-31
#1286 Fixes factionless players being able to access any locked block if they type /mf lock before interacting with it.

Mr-Deej:
Added a factionless check to the lock and unlock command
Possible solution to #1286


dmccoystephenson:
Fantastic, thank you!

dmccoystephenson (approved):

Issue #1284, opened 2022-01-28
Fix InvalidPluginException occurring w/ older versions of Java Runtime.


dmccoystephenson:
[06:34:00] [Server thread/ERROR]: Could not load 'plugins/Medieval-Factions-4.5.jar' in folder 'plugins'
org.bukkit.plugin.InvalidPluginException: java.lang.UnsupportedClassVersionError: dansplugins/factionsystem/MedievalFactions has been compiled by a more recent version of the Java Runtime (class file version 60.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at org.bukkit.plugin.java.JavaPluginLoader.loadPlugin(JavaPluginLoader.java:139) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at org.bukkit.plugin.SimplePluginManager.loadPlugin(SimplePluginManager.java:394) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at org.bukkit.plugin.SimplePluginManager.loadPlugins(SimplePluginManager.java:332) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at org.bukkit.craftbukkit.v1_16_R3.CraftServer.loadPlugins(CraftServer.java:383) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.DedicatedServer.init(DedicatedServer.java:185) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.w(MinecraftServer.java:809) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at net.minecraft.server.v1_16_R3.MinecraftServer.lambda$0(MinecraftServer.java:164) ~[spigot-1.16.5.jar:2991-Spigot-018b9a0-f3f3094]
	at java.lang.Thread.run(Thread.java:748) [?:1.8.0_292]

dmccoystephenson:
It should be noted that I ran into this error on 1.16.5.

VoChiDanh:
use java 16 and add `-DPaper.IgnoreJavaVersion=true` in cmd to run server. It will be ignore java version and run plugins

VoChiDanh:
i dont sure about spigot, i tried with paper and it work

[...]

PR #1304, opened 2022-01-31
Can ignore this :v: 

VoChiDanh:


.gitignore
```diff
@@ -0,0 +1,112 @@
+# User-specific stuff
```
dmccoystephenson, line 1: Is this necessary? Can you please remove this from the tracked files if it is not necessary?

.idea/.gitignore
```diff
@@ -0,0 +1,8 @@
+# Default ignored files
```
dmccoystephenson, line 1: Is this necessary? Can you please remove this from the tracked files if it is not necessary?

.idea/compiler.xml
```diff
@@ -0,0 +1,13 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dmccoystephenson, line 1: Is this necessary? Can you please remove this from the tracked files if it is not necessary?

.idea/encodings.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
dmccoystephenson, line 1: Is this necessary? Can you please remove this from the tracked files if it is not necessary?

dmccoystephenson (changes requested):
The changes look good! I left a few comments concerning whether or not some files were necessary to add to the tracked files.

.gitignore
VoChiDanh: This is just a small file to help you not push unnecessary folders and files, i think so

.idea/.gitignore
VoChiDanh: This is just a small file to help you not push unnecessary folders and files, i think so

VoChiDanh: This file came from gitignore, no big deal

.idea/compiler.xml
VoChiDanh: This file came from gitignore, no big deal

.idea/encodings.xml
VoChiDanh: This file came from gitignore, no big deal

dmccoystephenson (approved):

Mr-Deej:
this possibly closed #1284 will need feed back from players on 1.16 and java 8 or 11

Issue #1322, opened 2022-02-02
Fix players getting teleported to each other’s faction homes if they initiate the home command at the same time.


dmccoystephenson:
**Describe the bug**

Players are teleported to each other’s faction homes.

---
**To Reproduce**
Two players must initiate the /mf home command at the same time.

---
**Expected behavior**
The players should teleport to their respective faction homes.

dmccoystephenson:
A queuing system might be a potential solution to this problem. When implemented, players would be added to a queue when typing /mf home and would have to wait their turn before teleporting.

dmccoystephenson:
This has not been verified to be fixed, but it might be fixed.

alyphen:
This is the kind of issue caused by statefulness of commands, and is likely wider-reaching than solely home. All state should be removed from commands & listeners, and all required state should be kept within scope of the execution (i.e. scoped to onCommand in this case)

PR #1370, opened 2022-03-05
Alternative Implementation of Scheduler.scheduleTeleport()

dmccoystephenson:
This alternative implementation uses a custom class rather than the scheduler provided by the plugin.

This is intended to be a step in the right direction towards resolving #1322.

### DIFF
--- a/src/main/java/dansplugins/factionsystem/utils/extended/Scheduler.java
+++ b/src/main/java/dansplugins/factionsystem/utils/extended/Scheduler.java
@@ -121,11 +121,16 @@ public class Scheduler {
             try {
                 delay();
             } catch(Exception e) {
+                player.sendMessage(ChatColor.RED + "Something went wrong.");
                 Logger.getInstance().log("Something went wrong running a delayed teleport task.");
                 return;
             }
+
             if (playerHasNotMoved()) {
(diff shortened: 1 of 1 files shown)


VoChiDanh (commented):
Wow, this feels so new to me 🤔 , Thanks for requested me to review the code but I'm new to coding and I don't think I can fully appreciate a code. 😵

Mr-Deej (approved):
👍 
Only real note I have is wether the Delay is configurable amount of time, I see some people maybe wanting to change it to lower or longer

dmccoystephenson:
> 👍
> Only real note I have is wether the Delay is configurable amount of time, I see some people maybe wanting to change it to lower or longer

Thanks for the note! I recently made the switch to using a config option for this instead of hard coding it.

dmccoystephenson:
> Wow, this feels so new to me 🤔 , Thanks for requested me to review the code but I'm new to coding and I don't think I can fully appreciate a code. 😵

New or not, I still value your feedback. Don’t feel pressured to participate in reviews, though. Let me know if you’re open to them in the future.

dmccoystephenson:
These changes have introduced #1375

PR #1383, opened 2022-03-07
Locale Utility - Simplified Referencing Method

dmccoystephenson:
I've added a locale utility and simplified the referencing involved in making the plugin translatable.

### DIFF
--- a/src/main/java/dansplugins/factionsystem/utils/Locale.java
+++ b/src/main/java/dansplugins/factionsystem/utils/Locale.java
@@ -1,0 +1,9 @@
+package dansplugins.factionsystem.utils;
+
+import dansplugins.factionsystem.services.LocalLocaleService;
+
+public class Locale {
+    public static String get(String key) {
+        return Locale.get(key);
+    }
+}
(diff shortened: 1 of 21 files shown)


src/main/java/dansplugins/factionsystem/utils/Locale.java
```diff
@@ -0,0 +1,13 @@
+package dansplugins.factionsystem.utils;
+
+import dansplugins.factionsystem.services.LocalLocaleService;
+
+/**
+ * @author Daniel McCoy Stephenson
+ */
+public class Locale {
+    
+    public static String get(String key) {
```
VoChiDanh, line 10: It's still the same as if you left static in the file LocalLocaleService 🤔 , I think the problem is elsewhere, I will check it again after I get back from school

dmccoystephenson: This is just intended to make it easier to utilize the LocalLocaleService 👍 If we move to another solution it may be rendered unnecessary.

PR #1385, opened 2022-03-26
Fixing factionless players being unable to damage entities

### LINKED ISSUE #1353: Factionless players are unable to hit players that have joined one.
**Describe the bug**

---
**To Reproduce**

---
**Expected behavior**

---
**Screenshots or Logs**

---
**Java version**
 - [ ] Java 8
 - [ ] Java 11
 - [ ] Java 16
 - [ ] Java 17 
---
**Plugin version**
 - [ ] v4.1
 - [ ] v4.2
 - [ ] v4.3
 - [ ] v4.4
 - [ ] v4.5.x
---
**Spigot/Paper version** _(Please Specify which server software you are using)_
 - [ ] 1.16.x
 - [ ] 1.17.x
 - [ ] 1.18.x

dmccoystephenson:
After the PVP checks, all damage to entities caused by factionless players was being cancelled so as to prevent factionless players looting entities like armorstands. I've added a player check so that players are exempt from the cancellation.

closes #1353

### DIFF
--- a/src/main/java/dansplugins/factionsystem/eventhandlers/DamageHandler.java
+++ b/src/main/java/dansplugins/factionsystem/eventhandlers/DamageHandler.java
@@ -60,32 +60,42 @@ public class DamageHandler implements Listener {
      * 4) Players are not in the same faction but are not enemies.
      */
     private void handlePlayerVersusPlayer(Player attacker, Player victim, EntityDamageByEntityEvent event) {
+        Logger.getInstance().log("Handling damage between players.");
 
         // case 1
         if (arePlayersDueling(attacker, victim)) {
+            Logger.getInstance().log("Players are dueling. Ending if necessary.");
             endDuelIfNecessary(attacker, victim, event);
             return;
         }
 
         // case 2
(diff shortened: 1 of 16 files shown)


src/main/java/dansplugins/factionsystem/eventhandlers/DamageHandler.java
```diff
@@ -60,32 +60,42 @@ public void handle(EntityDamageByEntityEvent event) {
      * 4) Players are not in the same faction but are not enemies.
      */
     private void handlePlayerVersusPlayer(Player attacker, Player victim, EntityDamageByEntityEvent event) {
+        Logger.getInstance().log("Handling damage between players.");
```
VoChiDanh, line 63: i think we need add `debug: true` to config to prevent it spam when release

dmccoystephenson: Thanks for bringing this up! The method being called here actually only logs the message if the debug flag is enabled. I've made some changes to the Logger class to make this more clear.

PR #1386, opened 2022-03-27
Key Usage

dmccoystephenson:
These changes allow the Locale class to track the usage of keys. My hope is that this will be a step forward in allowing us to identify dead keys (keys that are no longer being used, but remain in our files).

### DIFF
--- a/src/main/java/dansplugins/factionsystem/utils/Locale.java
+++ b/src/main/java/dansplugins/factionsystem/utils/Locale.java
@@ -1,13 +1,29 @@
 package dansplugins.factionsystem.utils;
 
+import java.util.HashMap;
+
 import dansplugins.factionsystem.services.LocalLocaleService;
 
 /**
  * @author Daniel McCoy Stephenson
  */
 public class Locale {
+
+    private static HashMap<String, Integer> keyUses = new HashMap<>();
     
     public static String get(String key) {
+        recordKeyUse(key);
         return LocalLocaleService.getInstance().getText(key);
+    }
+
+    static void recordKeyUse(String key) {
+        if (keyUses.containsKey(key)) {
+            int numUses = keyUses.get(key);
+            keyUses.replace(key, numUses + 1);
+        }
+        else {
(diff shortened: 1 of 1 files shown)


src/main/java/dansplugins/factionsystem/utils/Locale.java
```diff
@@ -1,13 +1,29 @@
...
+public class Locale {
+
+    private static HashMap<String, Integer> keyUses = new HashMap<>();
+    
+    public static String get(String key) {
+        recordKeyUse(key);
+        return LocalLocaleService.getInstance().getText(key);
+    }
+
+    static void recordKeyUse(String key) {
```
VoChiDanh, line 19: I don't think this is necessary as it also partly depends on the user :thinking:

dmccoystephenson: If we track key usage and make it persistent, we'll be able to cycle through the keys loaded into the locale service and see which ones have 0 uses. Then we can created a report that lists the keys not being used and we can remove them from the .tsv files. There might be a better way of going about this, but I think this is a way forward.

VoChiDanh: then you can add metrics to make it easier to get data because if you put it here, only the person who gives the log will know

dmccoystephenson: I'm not familiar with Metrics. Would you be able to elaborate on this? @VoChiDanh

VoChiDanh: I haven't had time to research this yet

dmccoystephenson: No worries!

dmccoystephenson (commented):
At this time, no one can access key usage information that these changes allow us to save. Before merging, we should either make this information persistent and/or introduce a command to allow operators to view usage stats.

PR #1402, opened 2022-06-08
Convert to Gradle and other tweaks to build process

alyphen:
* Convert to gradle
* Relocate shaded dependencies
* Shade bstats rather than manually copying


dmccoystephenson (approved):
Looks good, thank you! That relocation should make things less error-prone, I hope.

PR #1412, opened 2022-06-26
Moving away from the Singleton pattern.

dmccoystephenson:
The changes in this PR are intended to force myself to actually manage my dependencies, rather than having an over-reliance on the singleton pattern.


dmccoystephenson:
Left to do:
- [ ] Deal with circular dependency between InetractionAccessChecker.java and PersistentData.java
- [ ] Refactor the commands

dmccoystephenson:
This is a depiction of the circular dependency. Some information may be missing.
circular dependency drawio

dmccoystephenson:
@alyphen
This is now ready for review. Let me know if you have any thoughts 👍

alyphen (approved):
Nice work on this, should make this project much easier to work with

PR #1413, opened 2022-07-01
Resolving Missing Dependencies

dmccoystephenson:
I missed some when transitioning from the singleton pattern.


dmccoystephenson:
**TO DO**
- [ ] Resolve circular dependency between ConfigService.java and LocaleService.java
- [ ] Resolve circular dependency between DynmapIntegrator.java and PersistentData.java
- [ ] Resolve circular dependency between BlockChecker.java and PersistentData.java

PR #1494, opened 2022-08-23
Add setting to toggle language file usage between tsv and yml and allow the "factionless" placeholder to be edited.

### LINKED ISSUE #1426: "Factionless" placeholder - editable
I just want, for everyone, to have an permission for editing this placeholder.

Clean command, or option in the config for editting the color, and text of this placeholder.

I dont have one. 

**Additional context**
I think this feature is pretty important for everyone, who wants to create some SMP server. 😄

VoChiDanh:
closes #1426


src/main/java/dansplugins/factionsystem/integrators/FiefsIntegrator.java
```diff
@@ -16,8 +16,8 @@ public class FiefsIntegrator {
 
     private FiefsAPI fi_api = null;
 
-    public FiefsIntegrator(MedievalFactions medievalFactions) {
```
alyphen, line 19: What's the purpose of this change?

src/main/java/dansplugins/factionsystem/placeholders/PlaceholderAPI.java
```diff
@@ -54,15 +55,15 @@ public boolean canRegister() {
...
+        id = id.toLowerCase(); // I'm unsure if PlaceholderAPI enforces case, but let's just do it to make sure.
         if (player == null) return null; // We only want to handle Player-Placeholders here.
 
         final boolean hasFaction = persistentData.isInFaction(player.getUniqueId());
         final Faction faction = persistentData.getPlayersFaction(player.getUniqueId());
 
         // Prerequisites.
         if (id.startsWith("faction_") && !hasFaction && !id.equalsIgnoreCase("faction_at_location")) {
-            return "Factionless"; // We don't want Faction-Specific Placeholders to return if they are Factionless!
+            return new ConfigService().getString("factionless"); // We don't want Faction-Specific Placeholders to return if they are Factionless!
```
alyphen, line 66: We should get the config service via dependency injection, not create a new instance each time we need it

src/main/java/dansplugins/factionsystem/services/MessageService.java
```diff
@@ -0,0 +1,44 @@
...
+import org.bukkit.configuration.file.FileConfiguration;
+import org.bukkit.configuration.file.YamlConfiguration;
+
+import java.io.File;
+import java.io.IOException;
+
+public class MessageService {
+
+
+    private static File languageFile;
```
alyphen, line 14: None of the fields or methods in this class need to be static

src/main/java/dansplugins/factionsystem/services/PlayerService.java
```diff
@@ -0,0 +1,47 @@
...
+
+    public static String getMessageType(String OldType, String NewType) {
+        if (MedievalFactions.USE_NEW_LANGUAGE_FILE) {
+            return NewType;
+        } else {
+            return OldType;
+        }
+    }
+
+    public static void sendMessageType(CommandSender sender, String OldType, String NewType, Boolean replace) {
```
alyphen, line 20: Variables should start with a lower case letter. replace can be a primitive boolean, it doesn't need the wrapper class, which allows null values to be passed in

src/main/java/dansplugins/factionsystem/services/PlayerService.java
```diff
@@ -0,0 +1,47 @@
...
+
+import dansplugins.factionsystem.MedievalFactions;
+import org.bukkit.ChatColor;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.ConsoleCommandSender;
+
+import java.util.List;
+
+public class PlayerService {
+
```
alyphen, line 11: None of the methods in this class need to be static.

alyphen (changes requested):

src/main/java/dansplugins/factionsystem/integrators/FiefsIntegrator.java
VoChiDanh: Looks like I made the wrong change

src/main/java/dansplugins/factionsystem/placeholders/PlaceholderAPI.java
VoChiDanh: This is probably the fastest and best way for users

src/main/java/dansplugins/factionsystem/services/PlayerService.java
VoChiDanh: replace here means, if true, it will get the message in new message, and if false then just type the first one of the message (I don't know what to call it but it's like getText(), I just type the one inside the quotes) then the method takes that line in the language file of yml to send it to the player, making typing a bit shorter

src/main/java/dansplugins/factionsystem/commands/MembersCommand.java
```diff
@@ -52,39 +57,72 @@ public void execute(CommandSender sender, String[] args, String key) {
...
-                        color = "&b";
-                    }
-                    if (faction.isOwner(player.getUniqueId())) {
-                        rank = "**";
-                        color = "&c";
-                    }
-                    return translate("&f" + player.getName() + color + rank);
-                }).forEach(sender::sendMessage);
-        sender.sendMessage(translate("&b----------\n"));
+        if (!new MedievalFactions().USE_NEW_LANGUAGE_FILE) {
```
alyphen, line 81: We don't want to create a new instance of the main plugin class - pass the reference into the command.

src/main/java/dansplugins/factionsystem/integrators/CurrenciesIntegrator.java
```diff
@@ -12,6 +12,6 @@
 public class CurrenciesIntegrator {
 
     public boolean isCurrenciesPresent() {
-        return (Bukkit.getServer().getPluginManager().getPlugin("Currencies") != null);
+        return Bukkit.getServer().getPluginManager().getPlugin("Currencies") == null;
```
alyphen, line 15: This inverts the logic! Surely if we want to check currencies _is_ present this should be returning that it isn't null?

src/main/java/dansplugins/factionsystem/integrators/DynmapIntegrator.java
```diff
@@ -404,7 +401,7 @@ else if ((ourblks != null) && (node.getWorldName().equalsIgnoreCase(curworld)) &
                 if (m == null) {
                     m = markerSet.createAreaMarker(polyid, name, false, curworld, x, z, false);
                     if (m == null) {
-                        System.out.println(String.format(localeService.get("ErrorAddingAreaMarker"), polyid));
+                        System.out.printf((localeService.get("ErrorAddingAreaMarker")) + "%n", polyid);
```
alyphen, line 404: Appreciate that the original code wasn't yours but since we're changing code in the area, perhaps we could use the plugin's logger rather than print statements?

src/main/java/dansplugins/factionsystem/objects/helper/FactionFlags.java
```diff
@@ -126,26 +127,27 @@ public void sendFlagList(Player player) {
...
         }
 
         if (flag.equals("fiefsEnabled") && !fiefsIntegrator.isFiefsPresent()) {
-            player.sendMessage("Fiefs either isn't enabled or present.");
+            new PlayerService().sendMessageType(player, "&cFiefs either isn't enabled or present.", "FiefsNotEnable", false);
             return;
         }
 
-        if (flag.equals("officersCanMintCurrency") && !currenciesIntegrator.isCurrenciesPresent()) {
+        if (flag.equals("officersCanMintCurrency") && currenciesIntegrator.isCurrenciesPresent()) {
```
alyphen, line 150: This, combined with the currencies integrator change, makes this more difficult to follow. We should be returning if currencies is not present, as the flag isn't supported.

src/main/java/dansplugins/factionsystem/services/CommandService.java
```diff
@@ -103,11 +103,19 @@ public boolean interpretCommand(CommandSender sender, String label, String[] arg
...
+                if (!medievalFactions.USE_NEW_LANGUAGE_FILE) {
+                    sender.sendMessage(ChatColor.AQUA + String.format(localeService.get("MedievalFactionsTitle"), medievalFactions.getVersion()));
+                    sender.sendMessage(ChatColor.AQUA + String.format(localeService.get("DeveloperList"), new MedievalFactions().getDescription().getAuthors()));
+                    sender.sendMessage(ChatColor.AQUA + localeService.get("WikiLink"));
+                    sender.sendMessage(ChatColor.AQUA + String.format(localeService.get("CurrentLanguageID"), configService.getString("languageid")));
+                    sender.sendMessage(ChatColor.AQUA + String.format(localeService.get("SupportedLanguageIDList"), localeService.getSupportedLanguageIDsSeparatedByCommas()));
+                } else {
+                    new MessageService().getLanguage().getStringList("PluginInfo")
+                            .forEach(s -> {
+                                s = s.replaceAll("#version#", medievalFactions.getVersion()).replaceAll("#dev#", medievalFactions.getDescription().getAuthors().toString());
```
alyphen, line 115: We probably want `replace` rather than `replaceAll` here - `replaceAll` does a regex replacement. Somewhat confusingly, `replace` in Java will replace all instances of a string.

src/main/java/dansplugins/factionsystem/services/ConfigService.java
```diff
@@ -163,7 +166,9 @@ public void handleVersionMismatch() {
         if (!getConfig().isInt("teleportDelay")) {
             getConfig().addDefault("teleportDelay", 3);
         }
-
+        if (!getConfig().isString("factionless")) {
+            getConfig().addDefault("factionless", "FactionLess");
```
alyphen, line 170: Perhaps `getConfig().set(...)` would be better here, or change the check to be `contains` rather than `isString`.
`addDefault` in combination with `options().copyDefaults()` will work if the key is not set, but that's not what we're checking - we're checking that the key _isn't a string_ - so in the case some other type is there, this will do nothing at all (as the default has already been overridden)

src/main/java/dansplugins/factionsystem/services/MessageService.java
```diff
@@ -0,0 +1,44 @@
...
+import java.io.IOException;
+
+public class MessageService {
+
+
+    private File languageFile;
+    private FileConfiguration language;
+
+    public void createLanguageFile() {
+        languageFile = new File(new MedievalFactions().getMedievalFactions().getDataFolder(), "language.yml");
```
alyphen, line 18: Pass the plugin instance into the message service.

src/main/java/dansplugins/factionsystem/services/MessageService.java
```diff
@@ -0,0 +1,44 @@
...
+        if (!languageFile.exists()) new MedievalFactions().getMedievalFactions().saveResource("language.yml", false);
+
+        language = new YamlConfiguration();
+        try {
+            language.load(languageFile);
+        } catch (IOException | InvalidConfigurationException e) {
+            e.printStackTrace();
```
alyphen, line 25: Perhaps use the plugin's logger to log exceptions.

src/main/java/dansplugins/factionsystem/services/PlayerService.java
```diff
@@ -0,0 +1,47 @@
...
+    public String getMessageType(String OldType, String NewType) {
```
alyphen, line 12: Variables should start with lowercase (i.e. `oldType` and `newType`)

alyphen (changes requested):

src/main/java/dansplugins/factionsystem/commands/AddLawCommand.java
```diff
@@ -44,13 +47,14 @@ public void execute(Player player, String[] args, String key) {
 
         // check if they have provided any strings beyond "addlaw"
         if (args.length == 0) {
-            player.sendMessage(translate("&c" + getText("UsageAddLaw")));
+            new PlayerService().sendMessageType(player, translate("&c" + getText("UsageAddLaw")), "UsageAddLaw", false);
```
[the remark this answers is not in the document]
VoChiDanh: What is dependency injection mean?

src/main/java/dansplugins/factionsystem/integrators/CurrenciesIntegrator.java
VoChiDanh: It will return false if plugin installed

src/main/java/dansplugins/factionsystem/integrators/DynmapIntegrator.java
VoChiDanh: I dont think i changed here, i think it automatic changed by reformat code

src/main/java/dansplugins/factionsystem/services/ConfigService.java
VoChiDanh: i follow the methods above, maybe i should change the whole method in there?

src/main/java/dansplugins/factionsystem/services/PlayerService.java
VoChiDanh: Oki, i'll change it in next push

src/main/java/dansplugins/factionsystem/commands/AddLawCommand.java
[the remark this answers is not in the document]
dmccoystephenson: If my understanding is correct, dependency injection is where any objects that a class depends on is passed through the constructor as parameters.

for example say that class A depends on class B - this is what class A would look like:

```
class A {
    b = null;

    A(B b) {
        this.b = b;
    }
}
```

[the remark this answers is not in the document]
alyphen: Dependency injection is a software design pattern in which the dependencies of a given service or component are passed into that service or component.
This means that services and components do not need to know the implementation details of each dependency, but is given everything it needs to function when it is created.
By building software in this way, we end up with loosely coupled components, which makes an application more flexible and easier to change parts later.

src/main/java/dansplugins/factionsystem/services/ConfigService.java
dmccoystephenson: This is an old habit of mine. We can replace any instances of .addDefault() with .set(), that's what I do with more recent projects now.

src/main/java/dansplugins/factionsystem/services/MessageService.java
dmccoystephenson: Yeah pass the main class to MessageService.java rather than instantiating a new instance here.

src/main/java/dansplugins/factionsystem/commands/AddLawCommand.java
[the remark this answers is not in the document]
VoChiDanh: oh I see. I'm not used to that style, I'll try anyway

src/main/java/dansplugins/factionsystem/integrators/CurrenciesIntegrator.java
alyphen: That runs counter to the method's name: `isCurrenciesPresent` should return whether currencies is present, not missing.
Either the change in behaviour should be reverted or the method should be renamed to make the intent clear.

VoChiDanh: oki, i'll change

dmccoystephenson:
closes #1426

src/main/java/dansplugins/factionsystem/commands/DuelCommand.java
```diff
@@ -39,88 +43,92 @@ public DuelCommand(LocaleService localeService, PersistentData persistentData, E
     public void execute(Player player, String[] args, String key) {
         final String permission = "mf.duel";
         if (!(checkPermissions(player, permission))) return;
-        if (args.length <= 0) {
+        if (args.length == 0) {
```
dmccoystephenson, line 46: Nice catch, thanks!

src/main/resources/en-us.tsv
```diff
@@ -1,551 +1,551 @@
...
-NoBlockDetectedToCheckForGate	No block detected to check for gate.
-NoFactionsDesignatedSingleQuotesRequired	No factions designated. Must be designated inside single quotes.
-NoLongerBypassingProtections	You are no longer bypassing protections provided by Medieval Factions.
-NoLongerInFactionChat	You are no longer in faction chat.
-NotAMemberOfAnyFaction	You are not a member of any faction.
-NotAnAlly	%s isn't an ally of yours.
-NotAnAllyOrVassal	%s isn't an ally or vassal of yours.
-NotAtWarWith	You aren't at war with %s
-NotAVassalOfAFaction	You aren't a vassal of a faction.
+AcceptedChallengeDuelHasBegun                   	You have accepted %s's challenge, the duel has begun.
```
dmccoystephenson, line 1: Do language key lookups still work with the added spaces here?

dmccoystephenson (approved):
This looks good to me! Thank you very much for your contribution. I'll let @alyphen take a final look before we merge this in.

src/main/resources/en-us.tsv
VoChiDanh: yes

src/main/java/dansplugins/factionsystem/commands/AllyCommand.java
```diff
@@ -85,25 +89,30 @@ public void execute(Player player, String[] args, String key) {
...
 
         // check if both factions are have requested an alliance
         if (faction.isRequestedAlly(otherFaction.getName()) && otherFaction.isRequestedAlly(faction.getName())) {
             // ally them
             faction.addAlly(otherFaction.getName());
             otherFaction.addAlly(faction.getName());
-
             // message player's faction
-            messageFaction(faction, translate("&a" + getText("AlertNowAlliedWith", otherFaction.getName())));
+            messageFaction(faction, translate("&a" + getText("AlertNowAlliedWith", otherFaction.getName())), Objects.requireNonNull(messageService.getLanguage().getString("AlertNowAlliedWith")).replaceAll("#faction#", otherFaction.getName()));
```
alyphen, line 112: use `replace` over `replaceAll` (`replaceAll` is for regex)

src/main/java/dansplugins/factionsystem/commands/CheckClaimCommand.java
```diff
@@ -39,9 +43,10 @@ public void execute(Player player, String[] args, String key) {
         final String result = chunkDataAccessor.checkOwnershipAtPlayerLocation(player);
 
         if (result.equals("unclaimed")) {
-            player.sendMessage(translate("&a" + getText("LandIsUnclaimed")));
+            playerService.sendMessageType(player, "&a" + getText("LandIsUnclaimed"), "LandIsUnclaimed", false);
         } else {
-            player.sendMessage(translate("&c" + getText("LandClaimedBy", result)));
+            playerService.sendMessageType(player, "&c" + getText("LandClaimedBy"), Objects.requireNonNull(messageService.getLanguage().getString("LandClaimedBy"))
+                    .replaceAll("#player#", result), true);
```
alyphen, line 49: Use `replace` over `replaceAll`

src/main/java/dansplugins/factionsystem/commands/CreateCommand.java
```diff
@@ -70,24 +76,30 @@ public void execute(Player player, String[] args, String key) {
...
         this.faction.addMember(player.getUniqueId());
 
         FactionCreateEvent createEvent = new FactionCreateEvent(this.faction, player);
         Bukkit.getPluginManager().callEvent(createEvent);
         if (!createEvent.isCancelled()) {
             persistentData.addFaction(this.faction);
-            player.sendMessage(translate("&a" + getText("FactionCreated")));
+            playerService.sendMessageType(player, "&a" + getText("FactionCreated"),
+                    Objects.requireNonNull(messageService.getLanguage().getString("FactionCreated"))
+                            .replaceAll("#name#", factionName), true);
```
alyphen, line 102: Use `replace` over `replaceAll`

src/main/java/dansplugins/factionsystem/commands/DeclareWarCommand.java
```diff
@@ -120,7 +136,10 @@ public void execute(Player player, String[] args, String key) {
             faction.addEnemy(opponent.getName());
             opponent.addEnemy(faction.getName());
             warFactory.createWar(faction, opponent);
-            messageServer(translate("&c" + getText("HasDeclaredWarAgainst", faction.getName(), opponent.getName())));
+            messageServer("&c" + getText("HasDeclaredWarAgainst", faction.getName(), opponent.getName()), Objects.requireNonNull(messageService.getLanguage().getString("HasDeclaredWarAgainst"))
+                    .replaceAll("#f_a#", faction.getName())
```
alyphen, line 140: Use `replace` over `replaceAll`

src/main/java/dansplugins/factionsystem/commands/DisbandCommand.java
```diff
@@ -57,13 +62,18 @@ public void execute(CommandSender sender, String[] args, String key) {
         if (args.length == 0) {
             if (!checkPermissions(sender, "mf.disband")) return;
             if (!(sender instanceof Player)) { // ONLY Players can be in a Faction
-                sender.sendMessage(translate(getText("OnlyPlayersCanUseCommand")));
+                if (!new MedievalFactions().USE_NEW_LANGUAGE_FILE) {
```
alyphen, line 65: Pass in the plugin instance rather than constructing a new one

src/main/java/dansplugins/factionsystem/commands/DisbandCommand.java
```diff
@@ -72,14 +82,20 @@ public void execute(CommandSender sender, String[] args, String key) {
             self = false;
         }
         if (disband == null) {
-            sender.sendMessage(translate("&c" + getText("FactionNotFound")));
+            playerService.sendMessageType(sender, "&c" + getText("FactionNotFound")
+                    , Objects.requireNonNull(messageService.getLanguage().getString("FactionNotFound"))
+                            .replaceAll("#faction#", String.join(" ", args)), true);
```
alyphen, line 87: Use `replace` over `replaceAll`

src/main/java/dansplugins/factionsystem/commands/GateCommand.java
```diff
@@ -42,29 +46,33 @@ public GateCommand(LocaleService localeService, PersistentData persistentData, E
     public void execute(Player player, String[] args, String key) {
         if (!(checkPermissions(player, "mf.gate"))) return;
         if (args.length == 0) {
-            player.sendMessage(translate("&b" + getText("SubCommands")));
-            player.sendMessage(translate("&b" + getText("HelpGateCreate")));
-            player.sendMessage(translate("&b" + getText("HelpGateName")));
-            player.sendMessage(translate("&b" + getText("HelpGateList")));
-            player.sendMessage(translate("&b" + getText("HelpGateRemove")));
-            player.sendMessage(translate("&b" + getText("HelpGateCancel")));
+            if (!new MedievalFactions().USE_NEW_LANGUAGE_FILE) {
```
alyphen, line 49: Pass in the plugin instance rather than constructing a new one

src/main/java/dansplugins/factionsystem/commands/GrantIndependenceCommand.java
```diff
@@ -51,9 +57,13 @@ public void execute(Player player, String[] args, String key) {
         target.setLiege("none");
         this.faction.removeVassal(target.getName());
         // inform all players in that faction that they are now independent
-        messageFaction(target, translate("&a" + getText("AlertGrantedIndependence", faction.getName())));
+        messageFaction(target, translate("&a" + getText("AlertGrantedIndependence", faction.getName())),
+                Objects.requireNonNull(messageService.getLanguage().getString("AlertGrantedIndependence"))
+                        .replaceAll("#name#", faction.getName()));
```
alyphen, line 62: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/GrantIndependenceCommand.java
```diff
@@ -51,9 +57,13 @@ public void execute(Player player, String[] args, String key) {
...
         // inform all players in players faction that a vassal was granted independence
-        messageFaction(faction, translate("&a" + getText("AlertNoLongerVassalFaction", target.getName())));
+        messageFaction(faction, translate("&a" + getText("AlertNoLongerVassalFaction", target.getName()))
+                , Objects.requireNonNull(messageService.getLanguage().getString("AlertNoLongerVassalFaction"))
+                        .replaceAll("#name#", target.getName()));
```
alyphen, line 66: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/InviteCommand.java
```diff
@@ -46,41 +49,51 @@ public InviteCommand(LocaleService localeService, PersistentData persistentData,
...
             return;
         }
         faction.invite(playerUUID);
         player.sendMessage(ChatColor.GREEN + localeService.get("InvitationSent"));
         if (target.isOnline() && target.getPlayer() != null) {
-            target.getPlayer().sendMessage(translate(
+            playerService.sendMessageType(target.getPlayer(),
                     "&a" + getText("AlertBeenInvited", faction.getName(), faction.getName())
-            ));
+                    , Objects.requireNonNull(messageService.getLanguage().getString("AlertBeenInvited")).replaceAll("#name#", faction.getName()),
```
alyphen, line 94: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/InviteCommand.java
```diff
@@ -89,9 +102,12 @@ public void execute(Player player, String[] args, String key) {
         getServer().getScheduler().runTaskLater(medievalFactions, () -> {
             faction.uninvite(playerUUID);
             if (tmp.isOnline() && tmp.getPlayer() != null) {
-                tmp.getPlayer().sendMessage(translate(
-                        "&c" + getText("InvitationExpired", faction.getName())
-                ));
+                playerService.sendMessageType(player,
+                        "&c" + getText("InvitationExpired", faction.getName()),
+                        Objects.requireNonNull(messageService.getLanguage().getString("InvitationExpired"))
+                                .replaceAll("#name#", faction.getName()),
```
alyphen, line 108: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/InvokeCommand.java
```diff
@@ -77,13 +83,13 @@ public void execute(Player player, String[] args, String key) {
             warringFaction.addEnemy(invokee.getName());
 
             messageFaction(invokee, // Message ally faction
-                    translate("&c" + getText("AlertCalledToWar1", faction.getName(), warringFaction.getName())));
+                    "&c" + getText("AlertCalledToWar1", faction.getName(), warringFaction.getName()), Objects.requireNonNull(messageService.getLanguage().getString("AlertCalledToWar1")).replaceAll("#f1#", faction.getName()).replaceAll("#f2#", warringFaction.getName()));
```
alyphen, line 86: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/InvokeCommand.java
```diff
@@ -77,13 +83,13 @@ public void execute(Player player, String[] args, String key) {
...
 
             messageFaction(warringFaction, // Message warring faction
-                    translate("&c" + getText("AlertCalledToWar2", faction.getName(), invokee.getName())));
+                    "&c" + getText("AlertCalledToWar2", faction.getName(), invokee.getName()), Objects.requireNonNull(messageService.getLanguage().getString("AlertCalledToWar2")).replaceAll("#f1#", faction.getName()).replaceAll("#f2#", invokee.getName()));
```
alyphen, line 89: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/InvokeCommand.java
```diff
@@ -77,13 +83,13 @@ public void execute(Player player, String[] args, String key) {
...
 
             messageFaction(this.faction, // Message player faction
-                    translate("&a" + getText("AlertCalledToWar3", invokee.getName(), warringFaction.getName())));
+                    "&a" + getText("AlertCalledToWar3", invokee.getName(), warringFaction.getName()), Objects.requireNonNull(messageService.getLanguage().getString("AlertCalledToWar3")).replaceAll("#f1#", faction.getName()).replaceAll("#f2#", warringFaction.getName()));
```
alyphen, line 92: Use replace over replaceAll

src/main/java/dansplugins/factionsystem/commands/StatsCommand.java
```diff
@@ -31,7 +34,18 @@ public void execute(Player player, String[] args, String key) {
 
     @Override
     public void execute(CommandSender sender, String[] args, String key) {
-        sender.sendMessage(ChatColor.AQUA + "=== Medieval Factions Stats ===");
-        sender.sendMessage(ChatColor.AQUA + "Number of factions: " + persistentData.getNumFactions());
+        if (!new MedievalFactions().USE_NEW_LANGUAGE_FILE) {
```
alyphen, line 37: Use plugin instance rather than creating a new one

alyphen (changes requested):
Services are looking much better, awesome work.

There's two minor flaws I can see in this still: usage of replaceAll where replace should be used (replaceAll should only be used for regex replacements) and creating new instances of the plugin class (this should never be done, pass the instance of the main class around rather than creating new ones)

I haven't commented on every single one since it's the same issue but I'm confident you can find them.

src/main/java/dansplugins/factionsystem/commands/AllyCommand.java
VoChiDanh: I use replaceAll because that can use number but Replace can't, however replaceAll still is better. This cannot be forced

VoChiDanh:
My machine just had a problem so maybe tomorrow morning I can double check if I'm using the instance somewhere else. Besides Using replace with replaceall, they are the same but replaceall can convert all to string and use it, and replace it is almost only used to change characters such as special characters... I will still keep as replaceall for my safety

alyphen:
The only difference between replace and replaceAll is that replaceAll uses a regular expression while replace uses plain strings.

src/main/java/dansplugins/factionsystem/commands/AllyCommand.java
dmccoystephenson: Try using replace and seeing if the behavior of the program changes.

dmccoystephenson:
> The only difference between replace and replaceAll is that replaceAll uses a regular expression while replace uses plain strings.

Could you clarify why it's preferable to replace plain strings rather than using a regular expression here?

src/main/java/dansplugins/factionsystem/commands/DisbandCommand.java
dmccoystephenson: Seconded.

alyphen:
Regular expressions are for pattern matching. When you want to match exact text you use a string, when you want to match a pattern like "two numbers followed by between 4 and 5 letters between a and f" or "any text surrounded by asterisks" or similar, that's when you use replaceAll (or Pattern if you need more advanced functionality)

src/main/java/dansplugins/factionsystem/commands/AllyCommand.java
dmccoystephenson: Any instances of .replaceAll have been replaced with .replace.

src/main/java/dansplugins/factionsystem/commands/DisbandCommand.java
dmccoystephenson: Done.

dmccoystephenson:
This is ready for re-review @alyphen

src/main/java/dansplugins/factionsystem/MedievalFactions.java
```diff
@@ -39,31 +38,45 @@
...
+    private final WarFactory warFactory = new WarFactory(persistentData);
+    private final RelationChecker relationChecker = new RelationChecker(persistentData);
+    private final GateService gateService = new GateService(persistentData, configService.getLocaleService(), ephemeralData, playerService, messageService);
+    private final LockService lockService = new LockService(persistentData, configService.getLocaleService(), persistentData.getBlockChecker(), playerService, messageService, ephemeralData);
+    private final Scheduler scheduler = new Scheduler(logger, configService.getLocaleService(), this, persistentData, configService, playerTeleporter, playerService, messageService);
+    private final CommandService commandService = new CommandService(configService.getLocaleService(), this, configService, persistentData, ephemeralData, persistentData.getChunkDataAccessor(), persistentData.getDynmapIntegrator(), warFactory, logger, scheduler, messenger, relationChecker, fiefsIntegrator, currenciesIntegrator, playerService, messageService);
+
+    public MedievalFactions getMedievalFactions() {
```
alyphen, line 64: This method isn't necessary

src/main/java/dansplugins/factionsystem/MedievalFactions.java
```diff
@@ -39,31 +38,45 @@
...
+    public ConfigService getConfigService() {
+        return configService;
+    }
 
     /**
      * This runs when the server starts.
      */
     @Override
     public void onEnable() {
+        medievalFactions = this;
```
alyphen, line 77: This isn't necessary

alyphen (changes requested):

src/main/java/dansplugins/factionsystem/MedievalFactions.java
dmccoystephenson: This has been removed.

src/main/java/dansplugins/factionsystem/MedievalFactions.java
dmccoystephenson: This has been removed.

dmccoystephenson:
This should be ready for re-review. @alyphen

alyphen (approved):

PR #1497, opened 2022-08-24
TabCompletion

jackcayc924:
Added the Tab Completion, let me know what you think.

Ty


src/main/java/dansplugins/factionsystem/MedievalFactions.java
```diff
@@ -69,6 +72,11 @@ public void onEnable() {
         registerEventHandlers();
         handleIntegrations();
         makeSureEveryPlayerExperiencesPowerDecay();
+
+        this.getCommand("mf").setTabCompleter(new TabCompleterBase(persistentData, configService));
```
alyphen, line 76: `this` can be omitted here

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
```diff
@@ -0,0 +1,588 @@
...
+import org.bukkit.entity.Player;
+
+import javax.annotation.Nullable;
+
+public class TabCompleterBase implements TabCompleter {
+
+	PersistentData persistentData;
+	ConfigService configService;
+	public TabCompleterBase(PersistentData persistentData, ConfigService configService)
+	{
```
alyphen, line 25: It would be good to maintain a consistent bracketing style with the rest of the project

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
```diff
@@ -0,0 +1,588 @@
...
+import org.bukkit.command.Command;
+import org.bukkit.command.CommandSender;
+import org.bukkit.command.TabCompleter;
+import org.bukkit.entity.Player;
+
+import javax.annotation.Nullable;
+
+public class TabCompleterBase implements TabCompleter {
+
+	PersistentData persistentData;
```
alyphen, line 22: These fields can be private

alyphen (changes requested):

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
```diff
@@ -0,0 +1,588 @@
...
+
+	public String removeBeginningAndEndQuotes(String targetedString)
+	{
+		return targetedString.replace("\"", "");
+	}
+
+	List<String> argsLength1 = new ArrayList<String>();
+
+	@Override
+	public List<String> onTabComplete(CommandSender sender, Command command, String alias, String[] args)
```
alyphen, line 70: This function is _huge_. Perhaps we could break it down into smaller functions (e.g. for subcommands)?

dmccoystephenson: Seconded. Splitting this up into more methods would be good.

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
dmccoystephenson: That would be preferable, yeah. We haven't actually defined the official bracketing style so this is something we should handle another time IMO.

dmccoystephenson (commented):
Thank you very much for your contribution! I imagine that players will be very excited about this new feature. After you get @alyphen's comments addressed I'd like to get this merged in.

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
dmccoystephenson: If it works, though, we don't necessarily have to clean this up in this PR.

src/main/java/dansplugins/factionsystem/MedievalFactions.java
dmccoystephenson: This has been omitted.

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
dmccoystephenson: These fields have been designated private.

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
dmccoystephenson: I've modified the bracketing style in this file.

src/main/java/dansplugins/factionsystem/commands/abs/TabCompleterBase.java
```diff
@@ -0,0 +1,474 @@
...
+
+	public String joinArgsBeyond(int index, String delim, String[] args) {
+		++index;
+		String[] data = new String[args.length - index];
+		System.arraycopy(args, index, data, 0, data.length);
+		return String.join(delim, data);
+	}
+
+	public List<String> filterStartingWith(String prefix, Stream<String> stream) {
+		return (List)stream.filter((s) -> {
```
alyphen, line 41: Cast should be unnecessary here

dmccoystephenson: This cast has been removed.

dmccoystephenson (approved):
Looks good to me! Thank you very much!

src/main/java/dansplugins/factionsystem/MedievalFactions.java
```diff
@@ -49,7 +51,7 @@ public class MedievalFactions extends PonderBukkitPlugin {
     private final PlayerTeleporter playerTeleporter = new PlayerTeleporter(logger);
     private final TerritoryOwnerNotifier territoryOwnerNotifier = new TerritoryOwnerNotifier(configService.getLocaleService(), configService, actionBarService);
     public boolean USE_NEW_LANGUAGE_FILE = configService.getBoolean("useNewLanguageFile");
-    private MedievalFactions medievalFactions;
+    private MedievalFactions medievalFactions = this;
```
alyphen, line 54: This field is unnecessary

src/main/java/dansplugins/factionsystem/MedievalFactions.java
```diff
@@ -49,7 +51,7 @@ public class MedievalFactions extends PonderBukkitPlugin {
...
     private final MessageService messageService = new MessageService(medievalFactions);
```
alyphen, line 55: We can just pass `this` into the services, we needn't use a reference within the class itself

alyphen (changes requested):

src/main/java/dansplugins/factionsystem/MedievalFactions.java
dmccoystephenson: We are now passing 'this' into the necessary services.

alyphen (approved):

Issue #1501, opened 2022-08-26
Fix error: "this.coord1" is null


ConfusedAlex:
## Describe the bug
Plugin cannot invoke dansplugins.factionsystem.objects.helper.GateCoord.getWorld() because this.coord1 is null

## To Reproduce
I dont know what to do to reproduce. This error gets from constantly.

## Expected behavior
No error

## Screenshots or Logs
```[06:00:59] [Server thread/WARN]: [MedievalFactions] Task #40 for MedievalFactions v4.7.0-SNAPSHOT generated an exception
java.lang.NullPointerException: Cannot invoke "dansplugins.factionsystem.objects.helper.GateCoord.getWorld()" because "this.coord1" is null
	at dansplugins.factionsystem.objects.domain.Gate.save(Gate.java:98) ~[Medieval-Factions-4.7.0-SNAPSHOT.jar:?]
	at dansplugins.factionsystem.objects.domain.Faction.save(Faction.java:410) ~[Medieval-Factions-4.7.0-SNAPSHOT.jar:?]
	at dansplugins.factionsystem.data.PersistentData$LocalStorageService.saveFactions(PersistentData.java:1351) ~[Medieval-Factions-4.7.0-SNAPSHOT.jar:?]
	at dansplugins.factionsystem.data.PersistentData$LocalStorageService.save(PersistentData.java:1328) ~[Medieval-Factions-4.7.0-SNAPSHOT.jar:?]
	at dansplugins.factionsystem.utils.extended.Scheduler$1.run(Scheduler.java:50) ~[Medieval-Factions-4.7.0-SNAPSHOT.jar:?]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftTask.run(CraftTask.java:101) ~[paper-1.19.2.jar:git-Paper-123]
	at org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftScheduler.mainThreadHeartbeat(CraftScheduler.java:483) ~[paper-1.19.2.jar:git-Paper-123]
	at net.minecraft.server.MinecraftServer.tickChildren(MinecraftServer.java:1492) ~[paper-1.19.2.jar:git-Paper-123]
	at net.minecraft.server.dedicated.DedicatedServer.tickChildren(DedicatedServer.java:446) ~[paper-1.19.2.jar:git-Paper-123]
[...]

PR #1502, opened 2022-08-26
Added NullCheck for coord1 and coord2

ConfusedAlex:
See  #1501


ConfusedAlex:
When one of those coords is null, should the whole gate be deleted?

alyphen:
We're having a broader discussion on how these get to be null and what the meaning of null is in the Discord.

If they _should_ be null, and that carries meaning, we need to add some logic around handling them when they're not present to the deserialization code also.

If they _shouldn't_ be null, we need to determine how they are getting to be null in the first place and fix the situation occurring.

dmccoystephenson:
Regardless, we should add logic around handling them when they're not present since it's possible for them to be null in older versions.

dmccoystephenson:
> When one of those coords is null, should the whole gate be deleted?

Maybe! Do you have any input on this? @philbgarner

src/main/java/dansplugins/factionsystem/objects/domain/Gate.java
```diff
@@ -95,9 +95,9 @@ public Map<String, String> save() {
         saveMap.put("open", String.valueOf(open));
         saveMap.put("vertical", String.valueOf(vertical));
         saveMap.put("material", material.name());
-        saveMap.put("world", coord1.getWorld());
-        saveMap.put("coord1", coord1.toString());
-        saveMap.put("coord2", coord2.toString());
+        if (coord1 != null) saveMap.put("world", coord1.getWorld());
```
dmccoystephenson, line 98: When loading gates, we should delete the gate if one of the coordinates is null.

dmccoystephenson: Let's also log this incident in the plugin's logger so that server owners are aware of why their gates disappeared.

dmccoystephenson (changes requested):
We can explore why the coordinates are null later. Let's focus on accounting for the possibility that it is null for now.

src/main/java/dansplugins/factionsystem/objects/domain/Gate.java
dmccoystephenson: Hey @ConfusedAlex! If you don't have time to address these comments would you mind giving me access to your fork? I'd love to get these fixes merged :)

ConfusedAlex: I've invited you :)

dmccoystephenson: This can be done another time. This resolves the problem for now. Thank you for your contribution.

dmccoystephenson: I opened an issue for this here: #1530

dmccoystephenson (approved):

PR #1519, opened 2022-09-16
Removing Fiefs & Currencies Integration + Startup Parsing Error Fix

dmccoystephenson:
## Description
All references to fiefs and currencies have been removed in order to resolve some circular dependencies.

## Startup Error Fix
I also fixed a startup parsing error, but now I'm experiencing this IO error (#1520) in the console upon executing commands. 

> [17:57:53 INFO]: ERROR: java.io.IOException: The system cannot find the path specified

This doesn't necessarily have to be resolved by this PR, as it isn't the focus of these changes. I've opened a new issue to track it here: https://github.com/Dans-Plugins/Medieval-Factions/issues/1520

## Reviewers
I've tagged a few community members as optional reviewers for this PR, but no one should feel pressured to give their input if they don't want to or don't have time. That being said, at least one approving review will be required to merge this PR.


dmccoystephenson:
It looks like the error is getting printed in the MessageService class on line 32. Here's the relevant code:

```
    public void createLanguageFile() {
        languageFile = new File(medievalFactions.getDataFolder(), "language.yml");
        if (!languageFile.exists()) {
            medievalFactions.saveResource("language.yml", false);
        }

        language = new YamlConfiguration();
        try {
            language.load(languageFile);
        } catch (IOException | InvalidConfigurationException e) {
            medievalFactions.getLogger().log(Level.WARNING, e.getCause().toString());
        }
    }
```

VoChiDanh:
I don't think that code is causing the above error...

VoChiDanh:
I don't think that code is causing the above error...

dmccoystephenson:
I think you're right @VoChiDanh now that I look at it again the code I posted involves a warning-level log message, rather than an info one in the error above.

src/main/resources/language.yml
```diff
@@ -93,7 +93,7 @@ CannotGrantAccessToSelf: "&cYou cannot grant access to yourself."
 RightClickGrantAccess: "&eRight click a chest or door to grant #name# access. Type /mf grantaccess cancel to cancel this."
 AlertGrantedIndependence: "&a#name# has granted your faction independence."
 AlertNoLongerVassalFaction: "&c#name# is no longer a vassal faction."
-FactionHomeNotSetYet: "&cThe faction home isn't set yet".
```
dmccoystephenson, line 96: This was the cause of the startup error, if anyone is curious.

VoChiDanh: Sorry about it. Can u use https://yamlchecker.com/ to check yml file if file have another typo?...

dmccoystephenson: No worries! It's valid 👍

image

dmccoystephenson (commented):
These changes look good to me. I'm going to go ahead and merge this PR.

PR #1521, opened 2022-09-23
PlayerService Renaming & Comments for Clarity

dmccoystephenson:
I've renamed the methods in the PlayerService class (and added comments) for clarity.

Please let me know if the new names and associated comments reflect your intentions with the class @VoChiDanh.

I added you also to see if you had any thoughts @alyphen.


VoChiDanh (approved):
nothing changed, just renamed the method for easy understanding but it's better than my method name. Thanks

Issue #1526, opened 2022-10-03
Fix #amount# and #pl# showing up in messages instead of the values that are meant to replace them.


IamJabbles:
## Describe the bug
(A clear and concise description of what the bug is.)
The bug involves to my knowledge only the #amount# and #pl# placeholder in the power gained line at the beginning of a day and /mf info.
## To Reproduce
(Steps to reproduce the behavior)
MF 4.7 default config
## Expected behavior
(A clear and concise description of what you expected to happen.)
To announce the amount of power gained (1) rather than #amount#
## Screenshots or Logs
(If applicable, add screenshots to help explain your problem.)

## Java version
 - [ ] Java 8
 - [ ] Java 11
 - [ ] Java 16
 - [X ] Java 17 

**Plugin version**
 - [ ] v4.1
 - [ ] v4.2
 - [ ] v4.3
 - [ ] v4.4
 - [ ] v4.5.x
 - [x] 4.7
---
**Spigot/Paper version** _(Please Specify which server software you are using)_
 - [ ] 1.16.x
 - [ ] 1.17.x
 - [x] 1.18.x

PR #1527, opened 2022-10-06
Fixed # showing up in a couple messages.

dmccoystephenson:
This includes:
- the scheduled power increase message
- the power shown in the output of the info command


dmccoystephenson:
This PR closed #1526

PR #1535, opened 2022-10-09
Fixed misleading translation

ConfusedAlex:
"aktiviert" translates into "activated". If you would toggle autoclaim off you would still get the feedback that autoclaim is now activated. This is misleading. The word "umgeschalten" fits better because it actually means toggling.


dmccoystephenson (approved):
That makes sense, thank you for the correction!

PR #1627, opened 2022-12-03
Feature - Add Member command

### LINKED ISSUE #1609: Add a command for forcefully adding players to a faction
Add a command, perhaps something like /f addmember, that forcefully adds a player to a faction.

dmccoystephenson:
The changes in this PR allow operators or those with permission to forcefully add members to factions.

closes #1609


dmccoystephenson:
German and French translations are required for the following:

CmdFactionAddMember=addmember
CommandFactionAddMemberNoPermission=You don't have permission to forcefully add members to factions.
CommandFactionAddMemberNotAPlayer=You must be a player to use this command.
CommandFactionAddMemberUsage=Usage: /mf addmember <player> <faction>
CommandFactionAddMemberInvalidTargetPlayer=That player hasn't played before.
CommandFactionAddMemberFailedToSavePlayer=Failed to save player.
CommandFactionAddMemberTargetPlayerAlreadyInFaction= That player is already in a faction.
CommandFactionAddMemberInvalidTargetFaction=That faction doesn't exist.
CommandFactionAddMemberFailedToSaveFaction=Failed to save faction.
CommandFactionAddMemberUnknownNewPlayerFaction=Unknown New Member
CommandFactionAddMemberSuccess=That player is now a member of the specified faction.
CommandFactionHelpAddMember=/faction addmember <player> <faction> - Forcefully add a member to a faction.

dmccoystephenson:
We just need french translations for the 'addmember' command now.

src/main/kotlin/com/dansplugins/factionsystem/command/faction/addmember/MfFactionAddMemberCommand.kt
```diff
@@ -0,0 +1,99 @@
...
+            sender.sendMessage("${ChatColor.RED}${plugin.language["CommandFactionAddMemberNotAPlayer"]}")
+            return true
+        }
+        if (args.isEmpty()) {
+            sender.sendMessage("${ChatColor.RED}${plugin.language["CommandFactionAddMemberUsage"]}")
+            return true
+        }
+
+        // get target player
+        val targetPlayer = plugin.server.getOfflinePlayer(args[0])
```
alyphen, line 33: I think we want the faction to be specified as well in this command, so we can add members to any other faction?

alyphen (changes requested):

src/main/kotlin/com/dansplugins/factionsystem/command/faction/addmember/MfFactionAddMemberCommand.kt
dmccoystephenson: The faction is specified on line 56.

alyphen: Ah I see! Maybe it would be better to have the faction as the first argument and use the last argument for player name?

dmccoystephenson: I view it almost like a sentence: Add <player> to <faction>

Also, if we switch the arguments around then we won't be able to easily join the arguments to create the faction name unless we remove the last argument first.

What do you think?

dmccoystephenson:
This is ready for re-review.

src/main/kotlin/com/dansplugins/factionsystem/command/faction/addmember/MfFactionAddMemberCommand.kt
alyphen: If it makes sense to you that's fine.
My general approach is to put things that vary the least first but that's definitely not a hard-and-fast rule.

src/main/resources/lang/lang_de_DE.properties
```diff
@@ -1017,6 +1017,18 @@ FactionDisbandedZeroPowerNotificationTitle=Fraktion aufgelï¿½st
 FactionDisbandedZeroPowerNotificationBody=Deine Fraktion, {0}, wurde aufgelï¿½st, da sie 0 Kraft erreicht hat.
 Teleporting=Teleportieren in {0} Sekunden...
 TeleportationCancelled=Teleportation abgebrochen.
-PowerIncreased=Du fï¿½hlst dich wieder stï¿½rker. Dein macht ist auf {0} gestiegen.
+PowerIncreased=Du f�hlst dich wieder st�rker. Dein macht ist auf {0} gestiegen.
```
alyphen, line 1020: Umlauts are displaying strangely for me here

dmccoystephenson: Yeah those look strange. I might just install linux tonight and re-commit some changes on there (assuming the problem is an OS issue)

PR #1629, opened 2022-12-05
Feature - Force Role Set

### LINKED ISSUE #1610: Allow /f role set to be used as admin with permission
/f role set should have an override for faction role permissions that allows admins to set roles

dmccoystephenson:
Players are now able to forcefully set members' roles in other factions if they are in bypass mode.

One translation key has been removed and another has been modified. The translations for the modified key need to be modified.

closes #1610


src/main/resources/lang/lang_de_DE.properties
```diff
@@ -49,9 +49,9 @@ CmdFactionFlagSet=setzen
 CmdFactionBypass=umgehen
 CmdFactionChat=chat
 CmdFactionChatHistory=geschichte
-CmdFactionClaimFill=ansprucherf�llen
+CmdFactionClaimFill=ansprucherf�llen
```
dmccoystephenson, line 52: The German and French translation files seem to have been messed up a bit. This needs to be resolved before this PR is merged into develop.

alyphen: Yep, accents need to be sorted out

PR #1632, opened 2022-12-09
JVM Name - MfFactionMember.getPlayerId()

### LINKED ISSUE #1595: Allow MfFactionMember.getPlayerId() to be called from Java
MfFactionMember.getPlayerId can currently not be called from Java due to the value class obfuscation.
Add a JvmName to it so we can get player IDs from Java for members.

dmccoystephenson:
This resolves #1595


src/main/kotlin/com/dansplugins/factionsystem/faction/MfFactionMember.kt
```diff
@@ -5,6 +5,7 @@ import com.dansplugins.factionsystem.player.MfPlayer
 import com.dansplugins.factionsystem.player.MfPlayerId
 
 data class MfFactionMember(
+    @get:JvmName("getId")
```
alyphen, line 8:
```suggestion
    @get:JvmName("getPlayerId")
```

alyphen (approved):

PR #1633, opened 2022-12-09
Flag - Allow Liege Interaction

### LINKED ISSUE #1612: Add a flag for allowing lieges to interact with land
Similar to the flag to allow vassals to interact with land, there should be one for lieges to interact with land.

dmccoystephenson:
The changes in this PR add a new flag, 'liegeChainCanInteractWithLand', which controls whether factions in the flagowner's liege chain can interact with the flagowner's land.

closes #1612


alyphen (changes requested):
Flag also needs to be added to [MfFlags.kt](https://github.com/Dans-Plugins/Medieval-Factions/blob/develop/src/main/kotlin/com/dansplugins/factionsystem/faction/flag/MfFlags.kt)

alyphen (approved):

PR #1643, opened 2022-12-13
Add linting

alyphen:
This PR adds linting. This will check code style on each build and add a pre-push hook to ensure code is linted.
Additionally, the linter has been run so some files' formatting has been changed.'


dmccoystephenson (approved):
Looks good, thanks for this!

PR #1656, opened 2022-12-15
Fixing a few bugs

Kyrenic:
Fixed some entries missing from the lang files, and a bug where people couldn't do /f claim fill because the permission was missing from the plugin.yml


dmccoystephenson (approved):
Looks good!

PR #1657, opened 2022-12-15
Added player permissions

Kyrenic:
I noticed that admins had an easy permission that could be given, I thought that having the same option for players would make using the plugin easier. It's default set to false to not override anything, but it'll still be an option for server owners to give in their permissions plugin.

### DIFF
--- a/src/main/resources/plugin.yml
+++ b/src/main/resources/plugin.yml
@@ -356,6 +356,70 @@ permissions:
     - mf.force.claim
     - mf.force.unclaim
     - mf.force.flag
+  mf.player:
+    description: Player permission
+    default: true
+    children:
+      - mf.help
+      - mf.ally
+      - mf.breakalliance
+      - mf.create
+      - mf.declarewar
+      - mf.invite
+      - mf.join
+      - mf.addlaw
+      - mf.laws
+      - mf.removelaw
+      - mf.makepeace
+      - mf.info
+      - mf.info.other
+      - mf.members
+      - mf.role.view
+      - mf.role.list
+      - mf.role.setpermission
+      - mf.role.set
+      - mf.role.create
+      - mf.role.delete
+      - mf.role.setdefault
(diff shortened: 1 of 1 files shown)


src/main/resources/plugin.yml
```diff
@@ -356,13 +356,77 @@ permissions:
     - mf.force.claim
     - mf.force.unclaim
     - mf.force.flag
+  mf.player:
+    description: Player permission
+    default: false
```
dmccoystephenson, line 361: Does this parent permission being false make the children permissions false too?

alyphen:
What's the use-case for this? Permissions should have sensible defaults so that players have the permissions they need by default.
The only case where I can see this being useful perhaps is greylisted servers

Kyrenic:
Well if you use a permissions plugin like for example LuckPerms, the default settings won't matter. In that case, you'd have to go through all the permissions to give players the proper perms to use the plugin. Given that the wiki isn't really updated (I think, let me know if the permissions on there are in fact updated), this would give server owners an easy way to give their players the permission to use the plugin.

src/main/resources/plugin.yml
Kyrenic: I don't think so, but I can certainly test it tonight

alyphen:
> Well if you use a permissions plugin like for example LuckPerms, the default settings won't matter. In that case, you'd have to go through all the permissions to give players the proper perms to use the plugin. Given that the wiki isn't really updated (I think, let me know if the permissions on there are in fact updated), this would give server owners an easy way to give their players the permission to use the plugin.

Permissions defaults are used unless there's an override by a permissions plugin, so they do still take effect.
Permissions plugins add permissions attachments to players which override the default permissions.
For permissions which default to true, all groups of whichever permissions plugin will have permission to use them unless there's an override to set them to false.

Kyrenic:
Ah, closing the issue then!

PR #1665, opened 2023-01-01
Fix character encoding for German language file

alyphen:
This PR fixes German character encoding.
The language files are encoded as ISO-8859-1 as required by the Properties specification:
https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Properties.html
> The [load(InputStream)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Properties.html#load(java.io.InputStream)) / [store(OutputStream, String)](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Properties.html#store(java.io.OutputStream,java.lang.String)) methods work the same way as the load(Reader)/store(Writer, String) pair, except the input/output stream is encoded in ISO 8859-1 character encoding. Characters that cannot be directly represented in this encoding can be written using Unicode escapes as defined in section [3.3](https://docs.oracle.com/javase/specs/jls/se17/html/jls-3.html#jls-3.3) of The Java Language Specification; only a single 'u' character is allowed in an escape sequence.


dmccoystephenson (approved):
Thank you for this!

Issue #1651, opened 2022-12-13
Allow people to view other factions laws


Kyrenic:
Would be nice to have, but not necessary.

Issue #1648, opened 2022-12-13
Don't display law delete button if player doesn't have permission to delete it


alyphen:
The Bukkit & faction permissions should be checked to see if someone can delete a law before displaying the X button.

PR #1673, opened 2023-01-05
Edits to law list, law remove, law edit and law move

Kyrenic:
This pull request should close two issues for milestone 5.1.0, namely #1651 and #1648 .


Kyrenic:
One thing to note is that we have to clear the law table, I currently see no way to give already existing laws a number index. If anyone has a suggestion, it would be very welcome, but for now I just opted to empty the table instead.

dmccoystephenson (commented):
Looks good! Let's just double check that the german character encoding is correct and shows up as intended in-game.

src/main/resources/lang/lang_de_DE.properties
```diff
@@ -157,7 +159,7 @@ CommandFactionCreateNameTooLong=Dein Fraktionsname darf nicht lï¿½nger als {0
 CommandFactionCreateFactionAlreadyExists=Eine Fraktion mit diesem Namen existiert bereits.
 CommandFactionCreateFactionFailedToSave=Ein Fehler ist beim Speichern deiner Fraktion aufgetreten.
 CommandFactionCreateSuccess=Fraktion {0} wurde erstellt.
-CommandFactionLawUsage=Benutze: /Fraktionsgesetz [Hinzufï¿½gen|Entfernen|Liste]
+CommandFactionLawUsage=Benutze: /Fraktionsgesetz [Hinzufï¿½gen|Entfernen|Liste|edit|move]
```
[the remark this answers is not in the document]
dmccoystephenson: Looks like there are some conflicts to resolve with the german language file. Resolving these might fix this.

src/main/kotlin/com/dansplugins/factionsystem/law/JooqMfLawRepository.kt
```diff
@@ -38,15 +48,56 @@ class JooqMfLawRepository(
     }
 
     override fun delete(id: MfLawId) {
+        delete(getLaw(id)!!)
```
alyphen, line 51: Remember that these operations can occur concurrently due to the async blocks - getLaw may not exist so asserting it does may cause errors.

src/main/kotlin/com/dansplugins/factionsystem/law/JooqMfLawRepository.kt
```diff
@@ -38,15 +48,56 @@ class JooqMfLawRepository(
...
+    }
+
+    override fun delete(law: MfLaw) {
         dsl.deleteFrom(MF_LAW)
-            .where(MF_LAW.ID.eq(id.value))
+            .where(MF_LAW.ID.eq(law.id.value))
+            .execute()
+        dsl.update(MF_LAW)
```
alyphen, line 58: Could this be achieved without updating all the numbers each time? Perhaps use it as a sort index for display and then treat the display number differently?

src/main/kotlin/com/dansplugins/factionsystem/law/JooqMfLawRepository.kt
```diff
@@ -38,15 +48,56 @@ class JooqMfLawRepository(
...
+    override fun edit(law: MfLaw, text: String) {
+        dsl.update(MF_LAW)
+            .set(MF_LAW.TEXT, text)
+            .where(MF_LAW.FACTION_ID.eq(law.factionId.value))
+            .and(MF_LAW.NUMBER.eq(law.number))
+            .execute()
+    }
+
+    override fun move(law: MfLaw, number: Int) {
+        if (law.number!! > number) {
```
alyphen, line 73: Please avoid `!!` if possible

src/main/kotlin/com/dansplugins/factionsystem/law/JooqMfLawRepository.kt
```diff
@@ -38,15 +48,56 @@ class JooqMfLawRepository(
...
-            .where(MF_LAW.ID.eq(id.value))
+            .where(MF_LAW.ID.eq(law.id.value))
+            .execute()
+        dsl.update(MF_LAW)
+            .set(MF_LAW.NUMBER, MF_LAW.NUMBER - 1)
+            .where(MF_LAW.FACTION_ID.eq(law.factionId.value))
+            .and(MF_LAW.NUMBER.gt(law.number))
+    }
+
+    override fun edit(law: MfLaw, text: String) {
```
alyphen, line 64: upsert can be used for this purpose - we don't need a separate query

src/main/resources/com/dansplugins/factionsystem/db/migration/V7__Store_law_index.sql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM `mf_law`;
+
+ALTER TABLE `mf_law` ADD `number` integer NOT NULL;
+
+--CREATE VIEW `lawView` AS SELECT ROW_NUMBER() OVER(PARTITION BY `faction_id`) AS `row_num`, `number` FROM `mf_law` ORDER BY `faction_id`, `row_num`
```
alyphen, line 5: Comments should be removed from this migration before merging.

src/main/resources/lang/lang_de_DE.properties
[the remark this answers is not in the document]
alyphen: I think this is caused by the base branch at the time this branch was created. It's worth doing a rebase.

alyphen (changes requested):

src/main/resources/com/dansplugins/factionsystem/db/migration/V7__Store_law_index.sql
```diff
@@ -0,0 +1,7 @@
+DELETE FROM `mf_law`;
+
+ALTER TABLE `mf_law` ADD `number` integer NOT NULL;
```
alyphen, line 3: Adding a default here or allowing null could prevent us having to empty the table.

alyphen:
> One thing to note is that we have to clear the law table, I currently see no way to give already existing laws a number index. If anyone has a suggestion, it would be very welcome, but for now I just opted to empty the table instead.

I don't think I'm comfortable emptying everybody's law table the moment this update is deployed. If we could assign null or a default value to the number/sort and then assign one during runtime when laws are viewed, that would be preferable I think.

src/main/kotlin/com/dansplugins/factionsystem/law/JooqMfLawRepository.kt
Kyrenic: I don't see how I would do that, but you're welcome to make a commit to my fork if you know how to.

Kyrenic:
The translation errors are still present on my branch, you'll need to walk me through how to rebase it since I don't know how to do that.

I've changed the SQL so it doesn't empty the table anymore, and it now assigns a number index when the laws are viewed if it doesn't have one already.

PR #1675, opened 2023-01-06
Fix - Unlock blocks upon getting broken

### LINKED ISSUE #1672: Fix blocks not getting unlocked upon breaking.
Blocks should get unlocked upon getting broken if they are locked.

dmccoystephenson:
Upon breaking a block, it will become unlocked if the breaker is the owner and it will be disallowed otherwise.

To avoid repetitive code, I've moved the 'unlock' method in the PlayerInteractListener class to the MfLockService class.

closes #1672


src/main/kotlin/com/dansplugins/factionsystem/locks/MfLockService.kt
```diff
@@ -43,6 +55,64 @@ class MfLockService(private val plugin: MedievalFactions, private val repository
         ServiceFailure(exception.toServiceFailureType(), "Service error: ${exception.message}", exception)
     }
 
+    public fun unlock(player: Player, block: Block) {
```
alyphen, line 58: Perhaps remove the player parameter and have this return some kind of result object with a failure message (either Result4k or its own type) - that would make this reusable beyond contexts which involve a Bukkit Player, and keeps view-related code out of the service.
If you use Result4k it also gives you the nice pattern of:
```kotlin
.onFailure { return it }
```
which lets you bubble up failures from methods you call

src/main/kotlin/com/dansplugins/factionsystem/listener/BlockBreakListener.kt
```diff
@@ -47,6 +47,18 @@ class BlockBreakListener(private val plugin: MedievalFactions) : Listener {
...
             }
         }
+
+        // if block locked and breaker is owner, unlock
+        val lockService = plugin.services.lockService
+        val lockedBlock = lockService.getLockedBlock(blockPosition) ?: return
+        if (lockedBlock.playerId == mfPlayer.id) {
+            lockService.unlock(event.player, event.block)
+        }
+        else {
```
alyphen, line 57: else should be on the same line as the closing brace

alyphen (changes requested):

src/main/kotlin/com/dansplugins/factionsystem/listener/BlockBreakListener.kt
dmccoystephenson: Fixed!

src/main/kotlin/com/dansplugins/factionsystem/locks/MfLockService.kt
dmccoystephenson: The player parameter has been removed and this method now returns a status indicator.

dmccoystephenson:
This is ready for re-review @alyphen

src/main/kotlin/com/dansplugins/factionsystem/locks/MfLockService.kt
```diff
@@ -43,6 +53,39 @@ class MfLockService(private val plugin: MedievalFactions, private val repository
         ServiceFailure(exception.toServiceFailureType(), "Service error: ${exception.message}", exception)
     }
 
+    fun unlock(block: Block): Int {
```
alyphen, line 56: Can we give this a more informative type rather than an int? We should avoid magic numbers where possible

dmccoystephenson: Yup! I've created and utilized the 'MfUnlockResult' enumeration instead.

dmccoystephenson: The push failed actually hold on I need to reinstall java.

dmccoystephenson: Okay changes have been pushed.

src/main/kotlin/com/dansplugins/factionsystem/locks/MfLockService.kt
```diff
@@ -43,6 +53,39 @@ class MfLockService(private val plugin: MedievalFactions, private val repository
...
+                toReturn = MfUnlockResult.NOT_LOCKED
+                return@Runnable
+            }
+            delete(lockedBlock.block).onFailure {
+                plugin.logger.log(SEVERE, "Failed to delete block: ${it.reason.message}", it.reason.cause)
+                toReturn = MfUnlockResult.FAILURE
+                return@Runnable
+            }
+        })
+        return toReturn
```
alyphen, line 88: This will basically always return SUCCESS because the asynchronous task will fire after returning.
Perhaps pass in a callback that receives the result?

dmccoystephenson:
Tested changes, resolved merge conflicts and formatted code.

alyphen (changes requested):

alyphen (approved):

PR #1676, opened 2023-01-06
Fix - Only Autoclaim With Permission

### LINKED ISSUE #1671: Fix players being able to auto claim even if their role doesn't have permission to claim.
When the owner toggles auto claim, players claim land even when their role doesn't grant them permission to do so.

dmccoystephenson:
This fixes players being able to autoclaim even if their role doesn't have the 'claim' permission.

closes #1671


alyphen (commented):
It may also be worth checking the normal/Bukkit permission in addition to the faction permission.
This would potentially allow for the use-case of claim prevention within WorldGuard regions, as LuckPerms can give players contextual permissions based on contexts with its ExtraContexts addon.

dmccoystephenson:
Good call! We're now checking the 'mf.claim' bukkit permission in addition to the player's role's permission.

dmccoystephenson:
This is ready for re-review.

alyphen (approved):

Issue #1628, opened 2022-12-04
MF Performance Optimizations


sujyrokimora:
## Describe the bug
When executing some tasks the server will crash or lag for a minute 
## To Reproduce
Die,mf claim, mf role, Power increase Event(anything that causes a dynmap update)

## Screenshots or Logs
```txt
[00:17:41 ERROR]: --- DO NOT REPORT THIS TO PAPER - THIS IS NOT A BUG OR A CRASH  - git-Paper-291 (MC: 1.19.2) ---
[00:17:41 ERROR]: The server has not responded for 10 seconds! Creating thread dump
[00:17:41 ERROR]: ------------------------------
[00:17:41 ERROR]: Server thread dump (Look for plugins here before reporting to Paper!):
[00:17:41 ERROR]: ------------------------------
[00:17:41 ERROR]: Current Thread: Server thread
[00:17:41 ERROR]:       PID: 29 | Suspended: false | Native: false | State: RUNNABLE
[00:17:41 ERROR]:       Stack:
[00:17:41 ERROR]:               java.base@17.0.2/java.util.HashMap.hash(HashMap.java:338)
[00:17:41 ERROR]:               java.base@17.0.2/java.util.HashMap.getNode(HashMap.java:568)
[00:17:41 ERROR]:               java.base@17.0.2/java.util.LinkedHashMap.get(LinkedHashMap.java:441)
[00:17:41 ERROR]:               medieval-factions-5.0.3-all.jar//com.dansplugins.factionsystem.dynmap.MfDynmapService.getPaths(MfDynmapService.kt:150)
[00:17:41 ERROR]:               medieval-factions-5.0.3-all.jar//com.dansplugins.factionsystem.dynmap.MfDynmapService.updateClaims(MfDynmapService.kt:85)
[00:17:41 ERROR]:               medieval-factions-5.0.3-all.jar//com.dansplugins.factionsystem.claim.MfClaimService.save$lambda-8$lambda-7(MfClaimService.kt:94)
[00:17:41 ERROR]:               medieval-factions-5.0.3-all.jar//com.dansplugins.factionsystem.claim.MfClaimService$$Lambda$13513/0x00000008033fa9b0.run(Unknown Source)
[00:17:41 ERROR]:               org.bukkit.craftbukkit.v1_19_R1.scheduler.CraftTask.run(CraftTask.java:101)
[...]

PR #1686, opened 2023-02-02
Fixed players being able to autoclaim land with no power restrictions.

### LINKED ISSUE #1685: Fix players being able to autoclaim land even if their faction doesn't have enough power to do so.
As of right now, players are able to claim land with the autoclaim functionality even if their faction is at or over its demesne limit. Players should be limited even in autoclaim mode.

dmccoystephenson:
closes #1685

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerMoveListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerMoveListener.kt
@@ -69,6 +69,21 @@ class PlayerMoveListener(private val plugin: MedievalFactions) : Listener {
                             )
                             return@Runnable
                         }
+
+                        // prevent autoclaiming if faction land >= max cumulative power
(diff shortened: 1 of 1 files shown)


alyphen:
This got caught in a merge conflict I think - I actually fixed it in the dynmap PR because the code is there already, it was just wrapped in the wrong if block

src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerMoveListener.kt
```diff
@@ -69,6 +69,21 @@ class PlayerMoveListener(private val plugin: MedievalFactions) : Listener {
                             )
                             return@Runnable
                         }
+
+                        // prevent autoclaiming if faction land >= max cumulative power
+                        if (plugin.config.getBoolean("factions.limitLand") && claimService.getClaims(playerFaction.id).size + 1 > playerFaction.power) {
```
alyphen, line 74: This is already present in the permission check around line 44 - you should just move it out of that block, we don't need to check twice.

dmccoystephenson:
Oh if this is already fixed in #1628 I'll just go ahead and close this one.

PR #1700, opened 2023-03-15
More Custom Charts

### LINKED ISSUE #1696: Start collecting data on average faction claims with bStats.
We should start collecting data on the average number of faction claims that servers have. This will help inform our decisions regarding optimization with dynmap among other parts of the plugin.

dmccoystephenson:
Added custom bStats charts for the following:
- database dialect
- average claims
- total claims
- initial power
- max power
- hours to reach max power
- hours to reach min power
- limit land
- allow neutrality

These can be seen on the bStats page.
https://bstats.org/plugin/bukkit/Medieval%20Factions/8929

closes #1696

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
@@ -248,7 +248,8 @@ class MedievalFactions : JavaPlugin() {
                 factionService.factions
                     .map {
                         claimService.getClaims(it.id).size
-                    }.average().toString()
+                    }
+                    .average().toInt().toString()
             }
         )
         metrics.addCustomChart(
@@ -284,7 +285,7 @@ class MedievalFactions : JavaPlugin() {
             }
         )
         metrics.addCustomChart(
-            SimplePie("allowNeutrality") {
+            SimplePie("allow_neutrality") {
                 config.getBoolean("factions.allowNeutrality").toString()
             }
         )
--- a/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
@@ -141,12 +141,6 @@ class MedievalFactions : JavaPlugin() {
         saveConfig()
 
         language = Language(this, config.getString("language") ?: "en-US")
-        val metrics = Metrics(this, 8929)
-        metrics.addCustomChart(
-            SimplePie("language_used") {
-                config.getString("language")
-            }
-        )
 
         Class.forName("org.h2.Driver")
         val hikariConfig = HikariConfig()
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
```diff
@@ -238,6 +232,64 @@ class MedievalFactions : JavaPlugin() {
...
+                config.getString("database.dialect")
+            }
+        )
+        metrics.addCustomChart(
+            SimplePie("average_claims") {
+                factionService.factions
+                    .map {
+                        claimService.getClaims(it.id).size
+                    }
+                    .average().toInt().toString()
```
alyphen, line 252: Does this need .toInt()? If it needs to be an int perhaps .roundToInt()?

src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
```diff
@@ -238,6 +232,64 @@ class MedievalFactions : JavaPlugin() {
...
+        metrics.addCustomChart(
+            SimplePie("database_dialect") {
+                config.getString("database.dialect")
+            }
+        )
+        metrics.addCustomChart(
+            SimplePie("average_claims") {
+                factionService.factions
+                    .map {
+                        claimService.getClaims(it.id).size
```
alyphen, line 250: I'm wondering whether it would be worth getting the database to do this work.

src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
```diff
@@ -238,6 +232,64 @@ class MedievalFactions : JavaPlugin() {
...
+                    }
+                    .average().toInt().toString()
+            }
+        )
+        metrics.addCustomChart(
+            SimplePie("total_claims") {
+                factionService.factions.sumOf {
+                    claimService.getClaims(it.id).size
```
alyphen, line 258: Again maybe worth doing on the DB? How does it perform with large numbers of factions and claims?

alyphen (changes requested):
Nice work - worth checking performance on some of these at large scales, and we probably don't want to be just truncating the decimal information from ints, but otherwise seems good.

src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
dmccoystephenson: Changed this to .roundToInt()!

src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
dmccoystephenson: Maybe! I don't imagine it will add too much strain to the server startup process but who knows?

src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
dmccoystephenson: Excellent question, no idea. My test server seems to start up pretty fast with the dev-generated test data though.

alyphen (approved):

PR #1703, opened 2023-03-19
Made players only steal power upon killing a player if the victim has…

### LINKED ISSUE #1661: Fix players gaining power upon killing even if their victim has no power.
Currently, you always gain power when you kill someone, even if the person you're killing has no power to lose themselves. This is potentially exploitable.

An config option to set a power steal value instead that takes power from the killed player and adds it to their power. The added benefit here is that a member at minimum power would not give power to their killer.

dmccoystephenson:
Players will no longer gain power when killing players with no power. Players with no power will also not lose any power.

closes #1661

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
@@ -26,6 +26,7 @@ class PlayerDeathListener(private val plugin: MedievalFactions) : Listener {
         val powerLostOnDeath = plugin.config.getDouble("players.powerLostOnDeath")
         val powerGainedOnKill = plugin.config.getDouble("players.powerGainedOnKill")
         val disbandZeroPowerFactions = plugin.config.getBoolean("factions.zeroPowerFactionsGetDisbanded")
+        val grantPowerToKillerIfVictimHasZeroPower = plugin.config.getBoolean("pvp.grantPowerToKillerIfVictimHasZeroPower")
 
         plugin.server.scheduler.runTaskAsynchronously(
             plugin,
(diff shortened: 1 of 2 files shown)


src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
```diff
@@ -35,31 +35,33 @@ class PlayerDeathListener(private val plugin: MedievalFactions) : Listener {
...
-                    addPowerToKiller(
-                        playerService,
-                        killer,
-                        killerMfPlayer,
-                        powerGainedOnKill
-                    )
-                    val killerFaction = factionService.getFaction(killerMfPlayer.id)
-                    if (disbandZeroPowerFactions && killerFaction != null && killerFaction.power <= 0.0) {
-                        disband(killerFaction, factionService)
+                    val killer = victim.killer ?: return@Runnable
```
alyphen, line 52: Can we make this configurable? I imagine some servers may wish to grant power for kills even if the person they are killing doesn't have any power.

src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
```diff
@@ -35,31 +35,33 @@ class PlayerDeathListener(private val plugin: MedievalFactions) : Listener {
...
+                    val killerMfPlayer = playerService.getPlayer(killer) ?: MfPlayer(plugin, killer)
+                    if (abs(powerGainedOnKill) > 0.00001) {
+                        addPowerToKiller(
+                            playerService,
+                            killer,
+                            killerMfPlayer,
+                            powerGainedOnKill
+                        )
+                        val killerFaction = factionService.getFaction(killerMfPlayer.id)
+                        if (disbandZeroPowerFactions && killerFaction != null && killerFaction.power <= 0.0) {
```
dmccoystephenson, line 62: Do we need to check for the killer's faction being 0 here? Most of the time (if not always) the killer will be gaining power upon a kill, so it would rarely be 0.

src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
dmccoystephenson: Yup good idea, I've added the `grantPowerToKillerIfVictimHasZeroPower` config option in the pvp section of the configuration file.

dmccoystephenson:
Converted this to a draft for now since my latest changes are untested.

src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerDeathListener.kt
alyphen: The case that comes to mind is negative power for the killer (i.e. punish people for killing people) - I don't imagine it's common but it's technically possible I think

dmccoystephenson: That makes sense!

dmccoystephenson:
I still haven't tested this but I think it'll work.

dmccoystephenson:
This has been verified to work on a 1.19.4 server.

alyphen (approved):

PR #1713, opened 2023-05-07
Added unique name check to `set name` command.

### LINKED ISSUE #1706: Two factions can share the same name.
Hello! I am loving the plugin and can't wait to see it in action on my server.

During testing, I noticed that two factions can share the same name.

This appears to break the /f declarewar command as well because when trying to enemy a faction that shares the same name as mine, I get: "There is no faction by that name".

Is there a way for me to disable the ability to create a faction using a name that is already taken?

Screenshot 2023-04-02 041802

dmccoystephenson:
We were checking for unique names when a faction was created, but not when it was renamed via the `set name `command. The changes in this PR add a unique name check to resolve this problem.

This was tested locally on a 1.18.2 spigot server.

closes #1706


alyphen (approved):
This seems adequate for most cases.

It may also be worth adding a unique constraint at the DB level in a migration, as if two players ran the command at exactly the same time, the query to update the faction is run async so both might succeed. That's absolutely an edge case though.

PR #1722, opened 2023-05-27
New Config Option - `factions.maxMembers`

### LINKED ISSUE #1691: Add a config option to limit the number of members a faction can have.
## Description
If this config option is set to 0, the functionality should be disabled.

## Acceptance Criteria
- [x] Add `factions.maxMembers` config option
- [x] Cancel invite command if the population of the sender's faction is greater than or equal to the `factions.maxMembers` config value.
- [x] Cancel join command if the population of the target faction is greater than or equal to the `factions.maxMembers` config value.
- [x] Cancel addmember command if the population of the target faction is greater than or equal to the `factions.maxMembers` config value.
- [x] Functionality verified on a test server.

dmccoystephenson:
The `factions.maxMembers` config option has been added.

closes #1691

### DIFF
--- a/src/main/resources/config.yml
+++ b/src/main/resources/config.yml
@@ -65,7 +65,7 @@ factions:
   allowNeutrality: false
   factionlessFactionName: Factionless
   factionHomeTeleportDelay: 5
-  maxMembers: 25
+  maxMembers: -1
   defaults:
     flags:
       alliesCanInteractWithLand: false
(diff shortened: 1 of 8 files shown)


src/main/resources/config.yml
```diff
@@ -65,6 +65,7 @@ factions:
   allowNeutrality: false
   factionlessFactionName: Factionless
   factionHomeTeleportDelay: 5
+  maxMembers: 25
```
alyphen, line 68: Perhaps we should default this to -1 so that servers currently running the plugin do not experience a change in behaviour when they upgrade?

alyphen (changes requested):

src/main/resources/config.yml
dmccoystephenson: Changed to -1!

alyphen (approved):

PR #1725, opened 2023-05-27
New Flag - `protectVillagerTrade`

### LINKED ISSUE #1541: Feature Request - protectVillagerTrade flag
A clear and concise description of what the problem is. 

My Server heavily revolves around the use of custom villagers, population, villager jobs etc. What I would love to see is a native solution within the MF4 plugin, is a flag to restrict players from other factions trading with another factions villagers. Much like the MobProtection flag but just cancelling the interaction event on Villagers instead.

Ideally: when protectVillagerTrade is True...
Player A in factions A is not allied with Player B in factions B. 
B comes into As territory. A has several hard to obtain Villagers that they have hired into their population. B tries to trade with the Villager but cannot because they are within As territory. If the Villager was in the Wilderness or the Villager was in Bs own or an allied territory, they could trade freely.

Vice versa, it would be unrestricted if set to false.

Main goal is to prevent non-allied players utilising the population of another faction.

I have tried assigning unique tags based on UUID to the villagers located within a certain radius of a marker (town center). However this Iis very clunky, works rarely and not ideal.

**Additional context**

Hrmmmm. Huh. Hrmmmm.

dmccoystephenson:
## Description
The `protectVillagerTrade` faction flag has been added to the plugin, allowing factions to control access to their villagers.

closes #1541 

## Testing
This has been tested on a 1.19.4 minecraft server. The flag can be set and trading behavior is as expected.

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractEntityListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractEntityListener.kt
@@ -1,0 +1,54 @@
+package com.dansplugins.factionsystem.listener
+
+import com.dansplugins.factionsystem.MedievalFactions
+import com.dansplugins.factionsystem.player.MfPlayer
+import dev.forkhandles.result4k.onFailure
+import org.bukkit.ChatColor
+import org.bukkit.entity.EntityType
+import org.bukkit.event.EventHandler
+import org.bukkit.event.Listener
+import org.bukkit.event.player.PlayerInteractEntityEvent
+import java.util.logging.Level
+
+class PlayerInteractEntityListener(private val plugin: MedievalFactions) : Listener {
+
+    @EventHandler
+    fun onPlayerInteractEntity(event: PlayerInteractEntityEvent) {
+        val playerService = plugin.services.playerService
+        val mfPlayer = playerService.getPlayer(event.player)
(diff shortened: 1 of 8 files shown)


src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractEntityListener.kt
```diff
@@ -0,0 +1,54 @@
...
+        if (clickedEntityType == EntityType.VILLAGER) {
+            val claimService = plugin.services.claimService
+            val claim = claimService.getClaim(clickedEntity.location.chunk) ?: return
+            val factionService = plugin.services.factionService
+            val claimFaction = factionService.getFaction(claim.factionId) ?: return
+
+            val protectVillagerFlagValue = claimFaction.flags[plugin.flags.protectVillagerTrade]
+            if (protectVillagerFlagValue && !claimService.isInteractionAllowed(mfPlayer.id, claim)) {
+                if (mfPlayer.isBypassEnabled && event.player.hasPermission("mf.bypass")) {
+                    event.player.sendMessage("${ChatColor.RED}${plugin.language["FactionTerritoryProtectionBypassed"]}")
```
alyphen, line 45: We may want to add a debounce to these at some point to prevent them getting to spammy - I'd regard that as out of scope for this PR though.

alyphen (approved):

PR #1726, opened 2023-05-27
New Config Option - `players.minPower`

### LINKED ISSUE #1660: Minimum power config option
Currently, players can only go down to 0 power, this potentially makes certain factions unconquerable if they have an alt that never goes into combat and loses power.

A config option to set the minimum amount of power a player can go down to. This would allow an owner to set a minimum amount of power a faction member can go down to, so if someone dies too much, they can actually be a detriment to the faction.

dmccoystephenson:
## Description
The `players.minPower` config option has been added, allowing players to have negative power.

closes #1660 

## Testing
This has been tested on a local 1.19.4 minecraft server.


src/main/kotlin/com/dansplugins/factionsystem/player/JooqMfPlayerRepository.kt
```diff
@@ -90,7 +91,7 @@ class JooqMfPlayerRepository(private val plugin: MedievalFactions, private val d
                 least(
                     value(maxPower),
                     greatest(
-                        value(0.0),
+                        value(minPower),
                         least(MF_PLAYER.POWER, MF_PLAYER.POWER_AT_LOGOUT, maxPower).div(MF_PLAYER.POWER_AT_LOGOUT)
```
alyphen, line 95: hoursToReachMin will no longer function as you expect it to - it changes the semantics of hoursToReachMin to be the time to reach 0 power.
The current formula for power is:
`power=((((time/hours to reach min)^4)*-1)+1)*power at logout`

In order to take the new minimum into account, you need to shift the graph downwards but also change the peak.
That means you want:
`power=(((((time/hours to reach min)^4)*-1)+1)*(power at logout - min power)) + min power`

The calculation run on the database does three things:
1. Runs the inverse of the power formula on the current power to determine the time we're at
2. Adds one hour to the calculated time
3. Runs the power formula on the new updated time to determine the new power.

Step (2) is the line written `.plus(timeIncrementHours)`, step (1) is everything before that, step (3) is everything after that.
Both the inverse and the normal formula need to include the extra operations to accomodate the new minimum power.

alyphen (changes requested):

src/main/kotlin/com/dansplugins/factionsystem/player/JooqMfPlayerRepository.kt
dmccoystephenson: Okay I think I've got it correct now. I'm a little rusty on inverses. Please let me know if it is correct now!

alyphen (approved):

dmccoystephenson (commented):
Looks good to me!

PR #1759, opened 2023-09-25
Resolve MariaDB driver when loading

alyphen:
To be tested by _M0nsters on Discord, who initially reported the issue.
This PR should force the MariaDB driver to load by resolving it, therefore running the static initializer block in its class which registers it.
This should be made obsolete by Java's service loader mechanism (we do use mergeServiceFiles() in the shadow plugin partially for this reason) but if people are still reporting it, there's no harm in having it there.


alyphen:
This doesn't fix the issue - _M0nsters seems to be using a modded server, I'm not sure if this affects the error.
We tried resolving MariaDB driver and also including a separate MySQL driver and neither fixes the class resolution.

PR #1762, opened 2023-10-24
Remove some unnecessary limitations on faction's land

Tems-py:


src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
```diff
@@ -161,8 +161,15 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
             if (mfPlayer.isBypassEnabled && event.player.hasPermission("mf.bypass")) {
                 event.player.sendMessage("$RED${plugin.language["FactionTerritoryProtectionBypassed"]}")
             } else {
-                event.isCancelled = true
-                event.player.sendMessage("$RED${plugin.language["CannotInteractWithBlockInFactionTerritory", claimFaction.name]}")
+                if (event.item?.type?.isEdible == true) {
+                    if (event.clickedBlock!!.type.isInteractable) {
```
alyphen, line 165: How do we know there is a clicked block? This may throw NPEs

Tems-py: > Represents an event that is called when a player interacts with an object or air
From spigot docs. I'm not sure but this means that there always will be block (air is also block?)

alyphen: Spigot includes nullability annotations, which mark fields as being nullable or not. In the vast majority of cases we should be able to avoid using '!!' and it's usually a sign that there's something wrong or going unhandled.

alyphen (approved):

alyphen:
All seems good! Please re-target to develop branch and then happy to approve

dmccoystephenson (commented):
Looks good to me! We should be able to merge once the target branch is set to 'develop'.

Thank you for your contribution!

alyphen (approved):

PR #1768, opened 2023-12-31
Config Option - Only Render Territories Upon Startup

### LINKED ISSUE #1767: Add config option for only updating the dynmap at server start
Updating the dynmap all of the time causes lag. One way to help this pain point would be to only update the dynmap upon server start.

A config option for only updating the dynmap upon server start should be added. If set to true, all updates except for the initial one will not be run.

No alternatives considered.

**Additional context**
Was mentioned as an alternative in #1766

dmccoystephenson:
## Problem
Rendering territories is causing a lot of lag for some servers.

## Solution
This PR introduces a config option to prevent the rendering of territories after the initial startup of the server. This limits the lag spike to one instance per session.

## Testing
This was used by a server owner in the community and their TPS improved from 15 to 18. This seems to ease the lag a bit.

## Related Issues
closes #1767

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/MedievalFactions.kt
@@ -300,6 +300,10 @@ class MedievalFactions : JavaPlugin() {
             MedievalFactionsPlaceholderExpansion(this).register()
         }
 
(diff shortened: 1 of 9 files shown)


alyphen (approved):
FlexTape
(It's important to remember this is very much a temporary fix and does not solve the problem at its source)

PR #1771, opened 2023-12-31
Config Option - Claim Fill Max Chunks

### LINKED ISSUE #1688: Add a configurable limit to claim fill chunks
It should be possible to limit the amount of chunks that claim fill can claim at once via the config.

dmccoystephenson:
## Problem
Currently there is no way to set a limit for the claim fill command. This can result in lagging out the server.

## Solution
A configurable limit called `claimFillMaxChunks` has been added as a config option. If set to a number above 0, attempts to use the claim fill command on a higher number of chunks than the limit will fail.

## Related Issue
closes #1688


src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt
```diff
@@ -89,6 +89,11 @@ class MfFactionClaimFillCommand(private val plugin: MedievalFactions) : CommandE
                             sender.sendMessage("$RED${plugin.language["CommandFactionClaimFillNoClaimableChunks"]}")
                             return@saveChunks
                         }
+                        val claimFillMaxChunks = plugin.config.getInt("factions.claimFillMaxChunks", -1)
```
alyphen, line 92: Bukkit API shouldn't be accessed asynchronously - this needs to be pulled out of the async task

src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt
```diff
@@ -89,6 +89,11 @@ class MfFactionClaimFillCommand(private val plugin: MedievalFactions) : CommandE
...
+                        if (claimFillMaxChunks > 0 && claimableChunks.size > claimFillMaxChunks) {
```
alyphen, line 93: Putting the check here will incur a performance penalty, as the `fill` function will still try to find more chunks than the cap. This check should go in `fill` instead and apply a circuit break as soon as it finds itself trying to find more chunks than the limit.

src/main/resources/lang/lang_de_DE.properties
```diff
@@ -1074,3 +1074,4 @@ CommandFactionJoinFactionFull=Die Fraktion, der du beitreten m
 CommandFactionAddMemberTargetFactionFull=Die Fraktion, zu der du diesen Spieler hinzuf�gen m�chtest, ist voll.
 PlayerInteractEntityFailedToSavePlayer=Der Spieler konnte nicht gespeichert werden.
 PlayerInteractEntityCannotTradeWithVillager=Du kannst nicht mit diesem Dorfbewohner handeln.
+CommandFactionClaimFillTooManyChunks=Sie k�nnen nicht mehr als {0} Chunks gleichzeitig beanspruchen.
```
alyphen, line 1077: Should use impolite form for consistency (Du kannst nicht mehr als {0} Chunks gleichzeitig beanspruchen.)

alyphen (changes requested):

src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt
dmccoystephenson: This call has been pulled out of the async task.

I feel like I've asked you this before but why shouldn't the Bukkit API be accessed asynchronously?

src/main/resources/lang/lang_de_DE.properties
dmccoystephenson: We're now using impolite form here :) Thanks for this!

src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt
dmccoystephenson: This check has been moved to the `fill` function. An exception is now thrown if it finds itself trying to find more chunks than the limit.

PR #1842, opened 2025-03-14
Fix Ally Placeholder Issue and Enable PlaceholderAPI Testing

dmccoystephenson:
## Problem  
The ally placeholder is currently not functioning, and testing is difficult because the test server does not have PlaceholderAPI installed.  

## Solution  
The test server was updated to support PlaceholderAPI integration. The ally placeholder issue was reproduced and successfully fixed.  

## Testing  
The test server was started, and the following command was verified to return a boolean:  

`/papi parse me %medievalfactions_faction_test_ally%`


src/test/kotlin/com/dansplugins/factionsystem/placeholder/MedievalFactionsPlaceholderExpansionTest.kt
```diff
@@ -0,0 +1,300 @@
...
+        val claimService = mock(MfClaimService::class.java)
+        val faction = mock(MfFaction::class.java)
+        val mfPlayer = mock(MfPlayer::class.java)
+
+        `when`(plugin.services.factionService).thenReturn(factionService)
+        `when`(plugin.services.playerService).thenReturn(playerService)
+        `when`(plugin.services.claimService).thenReturn(claimService)
+        `when`(playerService.getPlayer(player)).thenReturn(mfPlayer)
+        `when`(factionService.getFaction(mfPlayer.id)).thenReturn(faction)
+        `when`(claimService.getClaims(faction.id)).thenReturn(listOf())
```
dmccoystephenson, line 106: Each instance of a test involving an empty list and an expected value of zero should be accompanied by a test with a non-empty list and a non-zero expected value

dmccoystephenson: Addressed

PR #1853, opened 2025-03-23
Retry Gate Save on Optimistic Locking Failure

dmccoystephenson:
## Problem  
Optimistic locking failures occasionally occur when saving gates, leading to excessive console logs.  

## Solution  
If an optimistic locking failure occurs, the system will retry saving the gate.  

## Testing  
Gate persistence was verified on the test server by:  
1. Creating a gate.  
2. Shutting down the server.  
3. Restarting the server.  
4. Confirming that the gate still existed.

### DIFF
--- a/src/main/kotlin/com/dansplugins/factionsystem/gate/MfGateService.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/gate/MfGateService.kt
@@ -56,9 +56,21 @@ class MfGateService(
     fun getGatesByStatus(status: MfGateStatus) = gatesById.values.filter { it.status == status }
 
     fun save(gate: MfGate) = resultFrom {
-        val result = gateRepo.upsert(gate)
-        gatesById[result.id] = result
-        return@resultFrom result
+        try {
+            val result = gateRepo.upsert(gate)
+            gatesById[result.id] = result
+            return@resultFrom result
+        } catch (e: OptimisticLockingFailureException) {
+            plugin.logger.warning("Optimistic locking failure for gate with ID: ${gate.id}. Retrying...")
+            val reloadedGate = gateRepo.getGate(gate.id) ?: throw IllegalStateException("Gate not found: ${gate.id}")
+            val mergedGate = reloadedGate.copy(
+                status = gate.status,
+                version = reloadedGate.version + 1
+            )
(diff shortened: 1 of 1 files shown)


src/main/kotlin/com/dansplugins/factionsystem/gate/MfGateService.kt
```diff
@@ -56,9 +56,21 @@ class MfGateService(
...
+        try {
+            val result = gateRepo.upsert(gate)
+            gatesById[result.id] = result
+            return@resultFrom result
+        } catch (e: OptimisticLockingFailureException) {
+            plugin.logger.warning("Optimistic locking failure for gate with ID: ${gate.id}. Retrying...")
+            val reloadedGate = gateRepo.getGate(gate.id) ?: throw IllegalStateException("Gate not found: ${gate.id}")
+            val mergedGate = reloadedGate.copy(
+                status = gate.status,
+                version = reloadedGate.version + 1
```
renbinden, line 68: Version should be updated in the SQL so you shouldn't need to touch it in code: https://github.com/Dans-Plugins/Medieval-Factions/blob/main/src/main/kotlin/com/dansplugins/factionsystem/gate/JooqMfGateRepository.kt#L62

When you fetch the new gate you should have the most recent version so it should avoid the failure.

src/main/kotlin/com/dansplugins/factionsystem/gate/MfGateService.kt
```diff
@@ -56,9 +56,21 @@ class MfGateService(
...
-        return@resultFrom result
+        try {
+            val result = gateRepo.upsert(gate)
+            gatesById[result.id] = result
+            return@resultFrom result
+        } catch (e: OptimisticLockingFailureException) {
+            plugin.logger.warning("Optimistic locking failure for gate with ID: ${gate.id}. Retrying...")
+            val reloadedGate = gateRepo.getGate(gate.id) ?: throw IllegalStateException("Gate not found: ${gate.id}")
+            val mergedGate = reloadedGate.copy(
+                status = gate.status,
```
renbinden, line 67: There's a lot more than just status that could be updated - how do we know this is all that needs updating?

renbinden (changes requested):

PR #1938, opened 2026-01-03
Fix/windcharges

dmccoystephenson:


dmccoystephenson:
@copilot
Recent commits updated the linter version, which resulted in widespread formatting changes. To make reviews easier and reduce the risk of unintended breaking changes, formatting-only diffs should be kept to a minimum. Please review the recent commits, revert any unnecessary formatting changes, and ensure the linter configuration does not enforce broad new formatting rules.

PR #1974, opened 2026-06-08
 feat: upgrade Spigot API to 1.21.11 with backwards-compatible API changes

GomorrhaDev:
Upgrades the Spigot API dependency from  1.19.2-R0.1-SNAPSHOT  to  1.21.11-R0.1-SNAPSHOT  and adapts all code that
relied on API members that were renamed, removed, or changed since 1.19.2.

All changes maintain runtime backwards compatibility with servers running 1.17+ (as declared  api-version  in
plugin.yml ).

I also added Wind Charges from changing blocks in enemy faction territory (like in a now closed PR)


src/main/kotlin/com/dansplugins/factionsystem/utils/MfServerVersion.kt
```diff
@@ -0,0 +1,24 @@
...
+        return Triple(
+            parts.getOrNull(0)?.toIntOrNull() ?: 0,
+            parts.getOrNull(1)?.toIntOrNull() ?: 0,
+            parts.getOrNull(2)?.toIntOrNull() ?: 0
+        )
+    }
+
+    fun isAtLeast(major: Int, minor: Int = 0, patch: Int = 0): Boolean {
+        val (actualMajor, actualMinor, actualPatch) = parse()
+        if (actualMajor > major) return true
```
dmccoystephenson, line 29: nitpick (non-blocking): `parse()` result could be cached

`parse()` is called on every `isAtLeast()` invocation, which is triggered inside event handlers. Since the server version is never changed at runtime, the result could be cached lazily:

```kotlin
private val version: Triple<Int, Int, Int> by lazy { parse() }

fun isAtLeast(major: Int, minor: Int = 0, patch: Int = 0): Boolean {
    val (actualMajor, actualMinor, actualPatch) = version
    ...
}
```

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

src/main/kotlin/com/dansplugins/factionsystem/listener/AreaEffectCloudApplyListener.kt
```diff
@@ -11,21 +11,23 @@ import org.bukkit.potion.PotionEffectType
...
-        "HUNGER",
-        "POISON",
-        "SLOW",
-        "SLOW_DIGGING",
-        "UNLUCK",
-        "WEAKNESS",
-        "WITHER"
-    ).mapNotNull {
-        PotionEffectType.getByName(it)
+    private val harmfulPotionEffectTypes: List<PotionEffectType?> by lazy {
```
dmccoystephenson: nitpick (non-blocking): nullable type annotation could be corrected

`mapNotNull` is already used to filter out nulls, so the property type could be declared as `List<PotionEffectType>` rather than `List<PotionEffectType?>`. No nulls will ever be contained in the list at runtime, so the nullable annotation could be considered misleading.

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

src/main/kotlin/com/dansplugins/factionsystem/listener/EntityExplodeListener.kt
```diff
@@ -11,10 +14,30 @@ class EntityExplodeListener(private val plugin: MedievalFactions) : Listener {
...
+    private fun applyWindChargeProtection(event: EntityExplodeEvent) {
+        val entity = event.entity
+        if (entity.type.name != "WIND_CHARGE") return
+        val projectile = entity as? Projectile ?: return
+        val shooter = projectile.shooter as? Player ?: return
+        val playerService = plugin.services.playerService
+        val mfPlayer = playerService.getPlayer(shooter) ?: return
+        if (mfPlayer.isBypassEnabled && shooter.hasPermission("mf.bypass")) return
+        val claimService = plugin.services.claimService
+        val protectedBlocks = event.blockList().filter { block ->
```
dmccoystephenson, line 37: nitpick (non-blocking): intent of string-based type check could be documented

The string-based check (`entity.type.name != "WIND_CHARGE"`) is used rather than an `EntityType` enum reference because `EntityType.WIND_CHARGE` is not present in older API versions loaded at runtime. A short comment would be warranted here so the intent is not misread by a future contributor.

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

src/main/kotlin/com/dansplugins/factionsystem/listener/EntityDamageByEntityListener.kt
```diff
@@ -48,6 +48,10 @@ class EntityDamageByEntityListener(private val plugin: MedievalFactions) : Liste
             if (damagerDuel != null && damagedDuel != null && damagerDuel.id == damagedDuel.id) {
                 return
             }
+            if (damagerDuel != null || damagedDuel != null) {
```
dmccoystephenson, line 51: nitpick (non-blocking): new branch is not covered by a unit test

No test is provided for this new code path (attacker not in a duel, victim is in a duel → damage cancelled). The logic appears correct, but the intent would be better locked in if a unit test were added for the "one party in a duel, other not" scenario.

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

src/main/kotlin/com/dansplugins/factionsystem/utils/MfHostileMobChecker.kt
```diff
@@ -0,0 +1,19 @@
+package com.dansplugins.factionsystem.utils
+
+import org.bukkit.entity.Entity
+import org.bukkit.entity.Monster
+
+object MfHostileMobChecker {
+
+    fun isHostileMob(entity: Entity): Boolean {
```
dmccoystephenson, line 8: nitpick (non-blocking): version-branch logic is not directly tested

`MfHostileMobChecker` is only exercised indirectly through `CreatureSpawnListenerTest`. The version-branch logic would be made more explicit and resilient to future API changes if a small dedicated test were added — covering both the 1.19.3+ path and the pre-1.19.3 fallback with mocked `Entity` / `Monster` instances.

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

dmccoystephenson (commented):
Non-blocking notes from automated review.

dmccoystephenson (changes requested):
issue (blocking): CI build is failing — 11 unit tests are broken

The `build` check is currently red. The failures fall into four test classes, summarised below with root causes inferred from the stack traces.

---

**`AreaEffectCloudApplyListenerTest` — 2 failures**

- `onAreaEffectCloudApply_BasePotionDataIsNotHarmful_ShouldReturn` → `ExceptionInInitializerError` / `NullPointerException` at line 60
- `onAreaEffectCloudApply_BasePotionDataIsNull_ShouldReturnWithoutError` → `UnfinishedStubbingException` at line 35

The `by lazy` change causes `harmfulPotionEffectTypes` to be initialised on first access inside the test rather than at construction time. The existing test setup is not mocking `Bukkit.getBukkitVersion()` before the listener is exercised, so `MfServerVersion` (now called lazily) hits a null `Bukkit` instance. A `mockStatic(Bukkit::class.java)` block with a version stub will need to be added to this test class, matching the pattern used in `CreatureSpawnListenerTest`.

---

**`EntityExplodeListenerTest` — 5 failures**

- Most failures land at line 47 (`setUp`) with `MockitoException` / `UnfinishedVerificationException`
- `onEntityExplode_WindCharge_Pre121_ShouldNotCheck` → `NullPointerException` at line 115

The cascading `MockitoException` failures at `setUp` suggest that a previous test is leaving an open stubbing or an unclosed `MockedStatic`. The `onEntityExplode_WindCharge_Pre121_ShouldNotCheck` test casts `entity` to `Projectile` to stub `shooter`, but `entity` is created as a plain `Projectile` mock and the cast is unnecessary — the compiler warning ("No cast needed") already flags this. The NPE likely follows from the stubbing not being applied to the correct mock reference. These will need to be debugged and the mock lifecycle cleaned up.

---

**`InventoryMoveItemListenerTest` — 1 failure**

- `onInventoryMoveItem_DestinationIsUnlockedSingleChest_ShouldNotCancelEvent` → `InvalidUseOfMatchersException` at line 37

This appears to be a test pollution side-effect: an `UnfinishedStubbing` or stray `any()` matcher left open by an earlier failing test is bleeding into this class. Fixing the `PlayerInteractListenerTest` and `EntityExplodeListenerTest` issues above will likely resolve this as a by-product, but it should be confirmed after the other fixes land.

---

**`PlayerInteractListenerTest` — 3 failures**

- `onPlayerInteract_WartimeLadderPlacement_ConfigDisabled_ShouldCancel`
- `onPlayerInteract_WartimeLadderPlacement_ConfigEnabled_ShouldAllow`
- `onPlayerInteract_LadderInHand_PhysicalAction_ShouldNotBypass`

All three fail with `UnfinishedStubbingException` at line 1061, which is the `doReturn(true).when(itemMaterial).equals(Material.LADDER)` stub introduced in this PR. Mockito does not allow `equals()` to be stubbed on a mock this way — the method is used internally by the framework and overriding it produces an unfinished stubbing state. The `Material.LADDER` comparison in production code should be replicated in the test using the real `Material.LADDER` value rather than a mocked `Material` instance.

---

All four issues will need to be resolved before this PR can be merged.

*This comment was drafted by Claude on behalf of Daniel Stephenson.*

GomorrhaDev:
> issue (blocking): CI build is failing — 11 unit tests are broken
> 
> The `build` check is currently red. The failures fall into four test classes, summarised below with root causes inferred from the stack traces.
> 
> **`AreaEffectCloudApplyListenerTest` — 2 failures**
> 
> * `onAreaEffectCloudApply_BasePotionDataIsNotHarmful_ShouldReturn` → `ExceptionInInitializerError` / `NullPointerException` at line 60
> * `onAreaEffectCloudApply_BasePotionDataIsNull_ShouldReturnWithoutError` → `UnfinishedStubbingException` at line 35
> 
> The `by lazy` change causes `harmfulPotionEffectTypes` to be initialised on first access inside the test rather than at construction time. The existing test setup is not mocking `Bukkit.getBukkitVersion()` before the listener is exercised, so `MfServerVersion` (now called lazily) hits a null `Bukkit` instance. A `mockStatic(Bukkit::class.java)` block with a version stub will need to be added to this test class, matching the pattern used in `CreatureSpawnListenerTest`.
> 
> **`EntityExplodeListenerTest` — 5 failures**
> 
> * Most failures land at line 47 (`setUp`) with `MockitoException` / `UnfinishedVerificationException`
> * `onEntityExplode_WindCharge_Pre121_ShouldNotCheck` → `NullPointerException` at line 115
> 
> The cascading `MockitoException` failures at `setUp` suggest that a previous test is leaving an open stubbing or an unclosed `MockedStatic`. The `onEntityExplode_WindCharge_Pre121_ShouldNotCheck` test casts `entity` to `Projectile` to stub `shooter`, but `entity` is created as a plain `Projectile` mock and the cast is unnecessary — the compiler warning ("No cast needed") already flags this. The NPE likely follows from the stubbing not being applied to the correct mock reference. These will need to be debugged and the mock lifecycle cleaned up.
> 
> **`InventoryMoveItemListenerTest` — 1 failure**
> 
> * `onInventoryMoveItem_DestinationIsUnlockedSingleChest_ShouldNotCancelEvent` → `InvalidUseOfMatchersException` at line 37
> 
> This appears to be a test pollution side-effect: an `UnfinishedStubbing` or stray `any()` matcher left open by an earlier failing test is bleeding into this class. Fixing the `PlayerInteractListenerTest` and `EntityExplodeListenerTest` issues above will likely resolve this as a by-product, but it should be confirmed after the other fixes land.
> 
> **`PlayerInteractListenerTest` — 3 failures**
> 
> * `onPlayerInteract_WartimeLadderPlacement_ConfigDisabled_ShouldCancel`
> * `onPlayerInteract_WartimeLadderPlacement_ConfigEnabled_ShouldAllow`
> * `onPlayerInteract_LadderInHand_PhysicalAction_ShouldNotBypass`
> 
> All three fail with `UnfinishedStubbingException` at line 1061, which is the `doReturn(true).when(itemMaterial).equals(Material.LADDER)` stub introduced in this PR. Mockito does not allow `equals()` to be stubbed on a mock this way — the method is used internally by the framework and overriding it produces an unfinished stubbing state. The `Material.LADDER` comparison in production code should be replicated in the test using the real `Material.LADDER` value rather than a mocked `Material` instance.
> 
> All four issues will need to be resolved before this PR can be merged.
> 
> _This comment was drafted by Claude on behalf of Daniel Stephenson._

Currently working on tests

dmccoystephenson:
Sounds good. Thank you very much for your contribution!

dmccoystephenson:
## Follow-up review of the latest commits

**Verification:** The branch tip (`7254f20`) was checked out locally and the five affected test classes were run; a green build was observed. It should be noted that CI has not yet executed on this PR — the status rollup is currently empty and workflow runs appear to be awaiting maintainer approval — so the green result reflects a local run rather than CI.

### Blocking

**1. Ladder-placement coverage has been lost.** In `setupWartimeLadderTest`, `item.type` is now backed by a mocked `Material`, which can never be equal to the real `Material.LADDER` constant referenced at `PlayerInteractListener.kt:195`. As a result, `isPlacingLadder` is evaluated as `false` in every case, and the `ladderItem` parameter is no longer consumed (a `Parameter 'ladderItem' is never used` warning is emitted). The `WartimeLadderPlacement_*` and `LadderInHand_*` tests therefore pass without exercising the ladder branch they are named for. The real `Material.LADDER` constant should be used for `item.type` so that the discriminating behaviour is preserved.

**2. A change outside the stated scope appears to have been included.** At `EntityDamageByEntityListener.kt:51`, damage is now cancelled whenever either party is in a duel but not the same duel. This behaviour is unrelated to the API upgrade and wind-charge feature described in the PR, no test covers the new branch, and the effect — a dueling player can neither damage nor be damaged by anyone else — deserves explicit confirmation. It is recommended that this change be split into its own PR, or be covered by a test and justified.

### Worth confirming

**3. The definition of "hostile" has been broadened.** In `MfHostileMobChecker`, the `org.bukkit.entity.Enemy` interface (1.19.3+) is used in place of `Monster`. It was verified against the resolved 1.21.11 jar that `Monster extends Enemy`, while `Slime`, `Ghast`, `Phantom`, `Hoglin`, and `Shulker` implement `Enemy` without being `Monster`. Consequently, creature-spawn cancellation and mob protection are now applied to a larger set of entities. Confirmation that this expansion is intended would be appreciated.

### Minor

**4.** The "not harmful" assertion has been weakened. In `AreaEffectCloudApplyListenerTest`, the harmful list is emptied and a `null` effect type is returned, so the test would pass for any input and no longer distinguishes harmful from benign effects. A real harmful effect should be retained so that the filter is actually validated.

**5.** A test-only seam has been added to the production API: `harmfulPotionEffectTypeNames` is exposed as a public mutable `var` purely for injection. Restriction to `internal` (or a visible-for-testing annotation) would be preferable.

**6.** Only `WIND_CHARGE` is matched by name; breeze-fired `BREEZE_WIND_CHARGE` is not (the shooter-must-be-a-player guard already excludes it). It should simply be noted that only player-thrown charges are protected.

**7.** Some test cleanup is suggested, as flagged by the compiler: unused `chunk`/`block` parameters in `mockShooter`, an unused `shooter` local, several redundant `as Projectile` casts (`No cast needed`), and the unused `ladderItem` parameter.

### Acknowledged strengths

- Runtime backwards compatibility is preserved through reflection (`Enemy`), an `Attribute.valueOf` fallback, and version gating.
- `MfServerVersion` is cleanly implemented and well covered by tests.
- The global `versionProvider` is correctly saved and restored in `EntityExplodeListenerTest`, isolating the static state that the previous round left leaking.

**Recommendation:** Items 1 and 2 should be resolved before merge, and a decision on item 3 would be helpful; the remaining items are polish.

_This review was drafted by Claude on behalf of Daniel Stephenson._

dmccoystephenson:
**Re: item 3 (hostile-mob definition) — resolved, no change requested.**

On reflection, this change should be considered correct rather than left open. Both call sites were traced: in `CreatureSpawnListener` only hostile mobs are made eligible for spawn cancellation, and in `EntityDamageByEntityListener` hostile mobs are excluded from mob protection. Under the previous `Monster`-only check, `Slime`/`Ghast`/`Phantom`/`Hoglin` were mis-classified as passive, so their spawns were allowed through protected claims *and* they were shielded from being killed there. The `Enemy` interface captures the intended hostile set, so the switch is treated as a latent-bug fix and is endorsed.

Only items 1 (restore real `Material.LADDER` coverage) and 2 (the out-of-scope duel-damage change) are therefore considered outstanding.

_This note was drafted by Claude on behalf of Daniel Stephenson._

dmccoystephenson (approved):

Issue #1970, opened 2026-04-26
Refactor `PlayerInteractListener` to gate wartime permission checks strictly by Bukkit `Action` type


dmccoystephenson:
## Background

The wartime block permission system (`wartimePlaceableBlocks`, `wartimeBreakableBlocks`, `wartimeInteractableBlocks`) has produced multiple regressions because the listener conflates action type detection with wartime config evaluation. A right-click in Bukkit can mean either "place a block" or "interact with a block" depending on context, and the current listener infers which is happening while simultaneously evaluating wartime permissions — the two concerns are tangled, which is where leakage occurs (e.g. issue #1968, where holding a placeable block granted interaction rights).

## Goal

Strictly gate which wartime permission check is called based on the Bukkit `Action` enum value **before** any wartime config is consulted. The action type and the wartime permission lookup must be two completely separate steps with no overlap.

### In `PlayerInteractListener`

Restructure the handler so that action type is resolved first and is the sole determinant of which wartime check is invoked. The structure should follow this logic strictly:

- `Action.LEFT_CLICK_BLOCK` → only `isWartimeBreakableBlock` may be called. `isWartimePlaceableBlock` and `isWartimeInteractableBlock` must not be reachable from this branch.
- `Action.RIGHT_CLICK_BLOCK` where the clicked block is an interactive block (chest, lever, door, button, etc.) → only `isWartimeInteractableBlock` may be called. `isWartimePlaceableBlock` must not be reachable from this branch.
- `Action.RIGHT_CLICK_BLOCK` where the player is placing a block against a non-interactive surface → only `isWartimePlaceableBlock` may be called. `isWartimeInteractableBlock` must not be reachable from this branch.
- All other actions → no wartime checks are called.

The item held in the player's hand must play no role in determining which branch is taken. Branch selection must be based entirely on the `Action` and the type of the clicked block.

### Definition of "interactive block"

Introduce a well-defined predicate or constant set for what constitutes an interactive block — any block that responds to a right-click independently of what the player is holding (e.g. chests, furnaces, crafting tables, doors, trapdoors, fence gates, levers, buttons, repeaters, comparators, barrels, hoppers, droppers, dispensers, note blocks, jukeboxes, bells, lecterns, grindstones, looms, cartography tables, stonecutters, enchanting tables, anvils). This set should be defined once in a named constant or utility method, not inline in the listener, so it can be referenced in tests and updated in one place.

## Regression Tests to Add

[...]

Issue #1968, opened 2026-04-26
[Bug] Holding a block that is placeable in enemy territory allows you to interact with all blocks


ninjakitty844:
## Description
When holding a block that is in the `wartimePlaceableBlocks` list in the configuration file, you can interact with any block in enemy territory. You can open chests, flip levers, etc.

## Steps to Reproduce
*Provide step-by-step instructions to reproduce the issue.*
1. Put a ladder in the `wartimePlaceableBlocks` list in the Medieval Factions config
2. Declare war on another faction
3. Hold a ladder and interact with the other faction's blocks, it will work and you can interact with them

## Expected Behavior
The `wartimePlaceableBlocks` list is only meant to allow placement of these blocks, holding one of these blocks should not change which blocks you can now interact with.

### Java Version:
- [ ] Java 8
- [ ] Java 11
- [ ] Java 16
- [ ] Java 17
- [x] Java 23

### Plugin Version:
- [ ] v5.3.0
- [ ] v5.4.0
- [ ] v5.5.0
- [x] v5.8.0

### Server Software & Version:
- [ ] Spigot
- [x] Paper

Version: `1.19.4`

PR #1976, opened 2026-07-25
Stop pressure plates from flooding chat with protection messages

### LINKED ISSUE #1957: [Bug] Standing on a pressure plate can lag you with the number of "Cannot interact with blocks" messages
## Description
For every tick (1/20th of a second) you stand on a pressure plate that you cannot interact with (claimed by another faction that doesn't give you permission), you receive a message telling you that you cannot interact with it.

If I stand on TWO pressure plates at once, my computer can barely handle the number of messages it receives.

## Steps to Reproduce
1. Go into someone else's claims where you do not have permission to interact with blocks
2. Step on pressure plates

The bug is not critical, but my players tend to use pressure plates on top of fence posts to represent tables, and unfortunately my players drop frames any time they sit at a table like that.

## Expected Behavior
Maybe there should be a cooldown on "cannot interact" messages per-player?
Or maybe these messages just shouldn't show up for pressure plates?

On my server personally, I would prefer that pressure plates do not send a "cannot interact" message at all, since most of the time my players are not trying to interact with them even when stepping through them, and most players will understand that they cannot activate pressure plates in claims that they don't have permission to interact in.

## Actual Behavior
Image
It's hard to explain how much lag this causes, but for reference I get 120+ FPS when not standing on these pressure plates here. I have a chat mod that compresses these messages so I don't see all 100+ of them at once, the yellow number is how many times the same message has been repeated. I didn't stand on the pressure plates for long.

### Java Version:
- [ ] Java 8
- [ ] Java 11
- [ ] Java 16
- [ ] Java 17
- [x] Java 23

### Plugin Version:
- [ ] v5.3.0
- [ ] v5.4.0
- [ ] v5.5.0
- [x] v5.7.2

### Server Software & Version:
- [ ] Spigot
- [x] Paper

Version: `1.19.4`

dmccoystephenson:
## Summary

`PlayerInteractEvent` with `Action.PHYSICAL` (pressure plates, tripwire, farmland, turtle eggs) fires **once per tick** for as long as a player stands on the block. `PlayerInteractListener.applyProtections` sent a protection message on every one of those events, so a player stood on a pressure plate inside a claim they cannot interact with received ~20 messages/second — enough to drop the reporter from 120+ FPS to unplayable, and worse with several plates at once.

This suppresses the *message* for physical interactions only. **The interaction is still cancelled exactly as before** — no protection behaviour changes.

Suppressed at three sites, all reachable from a physical interaction:
- `CannotInteractWithBlockInFactionTerritory` (the reported case)
- `CannotInteractBlockInWilderness` (same flood when `wilderness.interaction.prevent` is on)
- `FactionTerritoryProtectionBypassed` (same flood for an admin with `mf.bypass` active)

## Test plan

- [x] `onPlayerInteract_PhysicalActionInFactionTerritory_ShouldCancelWithoutNotifyingPlayer` — cancelled, no message
- [x] `onPlayerInteract_RightClickInFactionTerritory_ShouldCancelAndNotifyPlayer` — contrast case, deliberate clicks are still explained
- [x] `onPlayerInteract_PhysicalActionInWilderness_...ShouldCancelWithoutNotifyingPlayer`
- [x] `onPlayerInteract_PhysicalActionWithBypassEnabled_ShouldNotNotifyPlayer`
- [x] `onPlayerInteract_LadderInHand_PhysicalAction_ShouldNotBypass` updated — it asserted a message *was* sent on a physical action; it now asserts the event is still cancelled and no message is sent. The protection assertion (`verifyEventCancelled`) is unchanged; only the message expectation moved with the intended behaviour.
- [ ] Manual: stand on a pressure plate in another faction's claim — no chat, plate still does not fire

## Verification anchor

**Local `./gradlew` could not run:** this checkout's only JDK is Java 21 (class file major 65) and the Gradle 7.6.4 wrapper rejects it (`Unsupported class file major version 65`); no Docker is available for the container fallback. Per the loop's UNVERIFIED rule, **the CI `build` check on this PR's head SHA is the anchor** — the changed files are squarely in CI's scope (Kotlin main + test sources), so the PR is not merge-ready until that check is green.

## Data-safety statement

No schema migrations; no DPC contract change; no `config.yml` change; no `plugin.yml` change; no `lang/` change (no new strings — the affected messages already exist and are still used for non-physical interactions).

## Deferred this cycle

Everything else in the open backlog was deferred because this cycle was scoped to a single well-bounded bug. Two things found while working here and worth their own cycle:

- **#1970** (`PlayerInteractListener` wartime `Action` gating) touches the same file but is a protection-semantics refactor that re-opens the #1968 exploit surface if resolved carelessly; it also asks for a curated "interactive block" set to replace Bukkit's `Material.isInteractable`, which is a real behaviour change. It deserves an undivided cycle.
- The `PlayerInteractListenerTest` fixture gives `block`, `block.getRelative(UP)` and `block.getRelative(DOWN)` all the same coordinates (`0,0,0`) in the same world, so they produce equal `MfBlockPosition` values. `setupLockedBlock` therefore has its `getLockedBlock` stub overwritten by its own `null` stub for the block above, and the locked-block tests never actually see a locked block. Filed separately; a physical-interaction test for the locked-block message path was dropped from this PR rather than built on that fixture.

Closes #1957

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 - bStats charts for DPC opt-in rate, login-reminder usage, server-IP sharing, and Discord-link presence.
 
 ### Fixed
+- Standing on a pressure plate (or tripwire, or farmland) you are not allowed to use no longer floods chat with a protection message every tick. Physical interactions are still blocked exactly as before; only the accompanying message is suppressed, for the faction-territory, wilderness and bypass notices alike.
 - Faction snapshot for the DPC sync is now collected on the Bukkit main thread before being dispatched off-thread via `HttpClient.sendAsync`. Off-thread access to `factionService.factions` could otherwise produce inconsistent reads or `ConcurrentModificationException` under load.
 - The DPC sync no longer POSTs an empty faction roster. A transient empty read (e.g. faction data not yet loaded at startup, or a reload mid-cycle) is skipped client-side rather than sent, so it can never depend on the provider's safety guards to avoid a faction wipe.
 
--- a/CONFIG.md
+++ b/CONFIG.md
@@ -123,7 +123,7 @@ Controls behavior and alerts in unclaimed wilderness areas.
 ### `wilderness.interaction.alert`
 **Type:** Boolean  
 **Default:** `false`  
-**Description:** When `true`, sends an alert message when players interact with blocks in wilderness.
+**Description:** When `true`, sends an alert message when players interact with blocks in wilderness. No alert is sent for physical interactions (stepping on a pressure plate or tripwire, trampling farmland), which repeat every tick and would flood chat; the interaction is still prevented.
 
 ### `wilderness.place.prevent`
 **Type:** Boolean  
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
@@ -107,6 +107,11 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
         val claimService = plugin.services.claimService
         val mfPlayer = playerService.getPlayer(event.player)
 
+        // Physical interactions (pressure plates, tripwire, farmland) fire once per tick for as long as
+        // the player is stood on the block, so notifying on each one floods the player's chat. The
+        // interaction is still cancelled - only the message is suppressed.
+        val suppressProtectionMessages = event.action == PHYSICAL
+
         if (mfPlayer == null) {
             event.isCancelled = true
             plugin.server.scheduler.runTaskAsynchronously(
@@ -182,7 +187,7 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
         if (claim == null) {
             if (plugin.config.getBoolean("wilderness.interaction.prevent", false)) {
                 event.isCancelled = true
-                if (plugin.config.getBoolean("wilderness.interaction.alert", true)) {
+                if (plugin.config.getBoolean("wilderness.interaction.alert", true) && !suppressProtectionMessages) {
                     event.player.sendMessage("$RED${plugin.language["CannotInteractBlockInWilderness"]}")
                 }
             }
@@ -201,7 +206,9 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
         // Check if player is allowed to interact based on faction relationships
         if (!claimService.isInteractionAllowed(mfPlayer.id, claim)) {
             if (mfPlayer.isBypassEnabled && event.player.hasPermission("mf.bypass")) {
-                event.player.sendMessage("$RED${plugin.language["FactionTerritoryProtectionBypassed"]}")
+                if (!suppressProtectionMessages) {
+                    event.player.sendMessage("$RED${plugin.language["FactionTerritoryProtectionBypassed"]}")
+                }
             } else {
                 // Check if player is at war and trying to place a ladder
                 // Only allow if they're right-clicking with a ladder on a solid, non-interactable block
@@ -234,7 +241,9 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
                     }
                 }
                 event.isCancelled = true
-                event.player.sendMessage("$RED${plugin.language["CannotInteractWithBlockInFactionTerritory", claimFaction.name]}")
+                if (!suppressProtectionMessages) {
+                    event.player.sendMessage("$RED${plugin.language["CannotInteractWithBlockInFactionTerritory", claimFaction.name]}")
+                }
             }
         }
     }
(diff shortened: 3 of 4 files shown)


dmccoystephenson:
Self-review rubric (Phase 4 — completed this cycle; the cycle that opened this PR was interrupted before posting it).

**External anchor:** `gh pr checks 1976` — `build` pass (3m7s), `docker-build` pass (6m50s) on head `48f3757e`. Local `./gradlew` still cannot run in this checkout: the wrapper starts (Gradle 7.6.4, JVM 21.0.10) but configuration dies with `BUG! exception in phase 'semantic analysis' in source unit '_BuildScript_' Unsupported class file major version 65` — Groovy in Gradle 7.6.4 cannot read Java 21 class files. No Docker for the container fallback. Per the loop's UNVERIFIED rule the CI check on the exact head SHA is the anchor, and the changed files (Kotlin main + test sources) are squarely inside CI's scope, so this is real verification rather than a coverage gap.

- **Scope:** PASS — 4 files, 121 insertions / 6 deletions. `PlayerInteractListener.kt` (+15/-6, the fix), `PlayerInteractListenerTest.kt` (+109, coverage), `CHANGELOG.md` (+1), `CONFIG.md` (1 line reworded). No unrelated formatting, renames or comment churn.
- **Tests-new:** PASS — no new public method is added; the changed private branch is covered by four new tests plus one contrast test (`RightClick...ShouldCancelAndNotifyPlayer`) that pins the non-physical path so the suppression cannot silently widen.
- **Tests-fix:** PASS — could not be run empirically (anchor unavailable locally), but the evidence is stronger than a stash-and-run here: the pre-existing test `onPlayerInteract_LadderInHand_PhysicalAction_ShouldNotBypass` asserted `verifyPlayerNotified()` on a `PHYSICAL` action and **had to be flipped** to `verifyPlayerNotNotified()` for this change. That is a recorded FAIL→PASS transition against the old behaviour, from a test written before the fix. Its `verifyEventCancelled()` assertion is untouched, so nothing about the protection itself was relaxed — only the message expectation moved with the intended behaviour change.
- **Sibling structure:** PASS — new tests follow the file's `Arrange / Act / Assert` comment blocks and its `onPlayerInteract_<Scenario>_Should<Outcome>` naming; the new `verifyPlayerNotNotified()` helper sits with `verifyPlayerNotified()` in the existing helper section.
- **Sibling renames:** PASS — no identifier renamed.
- **Docs:** PASS — full Phase 7 sweep run. `CHANGELOG.md` has an `[Unreleased] → Fixed` entry; `CONFIG.md` notes the physical-interaction exemption under `wilderness.interaction.alert` (the only documented option whose behaviour narrows). No command, faction flag, config option, dependency or schema change, so `COMMANDS.md` / `FACTION_FLAGS.md` / `DEPENDENCY_USAGE.md` / `DATABASE_QUERYING.md` are unaffected. `USER_GUIDE.md`, `README.md` and `README.pt-br.md` were grepped for `pressure plate` / `interaction.alert` / `cannot interact` — no hits, nothing to keep in sync.
- **Issue resolution:** PASS — `gh issue view 1957` re-read against the diff. The reported scenario (pressure plates on fence posts inside another faction's claim) routes to `CannotInteractWithBlockInFactionTerritory`, which is now suppressed. The reporter explicitly asked for "these messages just shouldn't show up for pressure plates" over a cooldown, which is what this implements.
- **Migrations-additive:** PASS — `git diff origin/main...HEAD -- '*db/migration/*'` is empty. No migration touched.
- **Lang-externalized:** PASS — no string literal added; all three affected messages are pre-existing `plugin.language[...]` keys and remain in use for non-physical interactions, so no `lang/` key is orphaned and no locale file needed editing.
- **DPC-contract:** PASS — nothing under `dpc/` or `pacts/` in the diff.
- **Main-thread-safety:** PASS — no new Bukkit API access and no new async dispatch; the change only adds boolean guards around existing synchronous `sendMessage` calls on the event thread.
- **Plugin.yml:** PASS — no new command or permission.
- **Config-defaults:** PASS — `src/main/resources/config.yml` is not in the diff. Only the root `CONFIG.md` prose changed; every default is unchanged, so no existing install's behaviour shifts on upgrade.

**Judgment call left for the reviewer (FAIL on completeness, not on correctness):** the fix does not cover the locked-block branch earlier in the same function. `LockProtectionBypassed` (`PlayerInteractListener.kt:158`) and `BlockLocked` (`:166`) are reachable from a `PHYSICAL` interaction — `lock()` applies no block-type restriction, so a pressure plate or farmland block can be `/mf lock`ed — and there they are worse than the case fixed here, because each per-tick message is emitted from a `runTaskAsynchronously` block that does a `playerService.getPlayer(...)` lookup. I deliberately did not extend the fix in this PR: it cannot be given regression coverage until #1977 is fixed (`setupLockedBlock` stubs `getLockedBlock` for the block and then for `getRelative(UP)`, which resolves to the same `(0,0,0)` position, so the second stub overwrites the first with `null` and no existing locked-block test ever sees a locked block), and `medievalFactions.server` is never mocked in this test class, so the branch would NPE the moment a test reached it. Filed as #1978 with the suggested fix. Adding an untested production change here — with no local anchor to run it against — was the worse trade.

**Data-safety statement:** no schema migration, no DPC contract change, no `config.yml` default change, no `plugin.yml` change, no `lang/` change. The do-not-auto-merge path check over `git diff --name-only origin/main...HEAD` matches nothing, and no single file loses more than 50 lines (6 deletions total, repo-wide).

**Deferred backlog:** this cycle was scoped to finishing this PR, so the rest of the open backlog is untouched. #1970 (wartime `Action` gating in this same file) remains the notable neighbour — it is a protection-semantics refactor that re-opens the #1968 exploit surface if resolved carelessly and deserves its own cycle.

---

Issue #1947, opened 2026-03-07
Add GitHub Actions workflow to build and deploy plugin to OMCSI


dmccoystephenson:
## Summary

Add a `.github/workflows/deploy.yml` file that automatically builds the plugin JAR and deploys it to the [omcsi (open-mc-server-infrastructure)](https://github.com/dmccoystephenson/open-mc-server-infrastructure) instance when changes are pushed to the deploy branch.

## Implementation

Create `.github/workflows/deploy.yml` modeled after the [Activity Tracker reference workflow](https://github.com/Dans-Plugins/Activity-Tracker/blob/main/.github/workflows/deploy.yml) with the following adjustments for this project:

- **JDK version:** `[8 / 11 / 17 / 21]`
- **Build tool & command:**
  - Maven: `mvn --batch-mode package`
  - Gradle: `./gradlew build`
- **JAR location:**
  - Maven: `target/` (exclude `original-*`, `*-sources.jar`, `*-javadoc.jar`)
  - Gradle: `build/libs/` (exclude `*-sources.jar`, `*-javadoc.jar`)

## Required Secrets

Add via repo Settings → Secrets and variables → Actions:

| Secret | Value |
|---|---|
| `OMCSI_DEPLOY_URL` | Base URL of the omcsi instance, e.g. `https://mc.example.com:8092` |
| `OMCSI_DEPLOY_TOKEN` | Value of `DEPLOY_AUTH_TOKEN` from the omcsi `.env` file |

## Required Variables

Add via repo Settings → Secrets and variables → Actions → Variables:

| Variable | Value |
|---|---|
| `PLUGIN_JAR_NAME` | Filename of the JAR on the server, e.g. `MyPlugin.jar` |
| `DEPLOY_BRANCH` | Branch that triggers deploys (default: `main`) |

## Acceptance Criteria

- [ ] `.github/workflows/deploy.yml` exists in the repo
- [ ] Pushing to the deploy branch triggers a build and successful deploy
- [ ] Workflow can also be triggered manually via the Actions tab (`workflow_dispatch`)
- [ ] Failed builds, missing variables, auth errors, and bad requests all produce a clear error message and non-zero exit code
- [ ] Secrets and variables documented in this ticket are configured in the repo

## Reference

[...]

Issue #1943, opened 2026-01-11
Implement JSON storage alternative


dmccoystephenson:
As a user, it would be convenient to not have to rely on a database in order to persist data for the plugin. An alternative persistence approach using JSON serialization and storage in the MedievalFactions directory should be implemented. Persistence logic should be refactored to be agnostic about the type of storage. JSON schemas should be used to validate data before storing or loading.

dmccoystephenson:
Resolved in PR #1944 (merged 2026-08-10): `storage.type: json` selects a JSON backend under `storage.json.path`, with one `Json*Repository` per entity type (12 in `storage/json/`), JSON schemas for `factions.json` and `players.json` under `src/main/resources/schemas/`, and `/faction migrate toJson` / `toDatabase` for moving between backends (`docs/MIGRATION_GUIDE.md`).

One part of the original ask is not fully met and is tracked separately in #2032: schema validation currently runs only on *write*, and only for `factions.json` and `players.json`; the other ten entity files have no schema, and nothing is validated when a file is read back.

This comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

_drafted by Claude on behalf of Daniel Stephenson_

Issue #1914, opened 2025-12-12
Expose functionality via REST API


dmccoystephenson:
Functionality intended to be used by other plugins should be exposed in a REST API. An OpenAPI spec should be generated from this REST API implementation so that other plugins can easily integrate with Medieval Factions. Unit tests should be implemented to verify that controllers forward requests to mocks of the expected components. A markdown document should be drafted showing an example use case (i.e. how a plugin might use the REST API to expand the functionality)

PR #1980, opened 2026-07-26
Stop locked blocks flooding chat on physical interactions

### LINKED ISSUE #1977: PlayerInteractListenerTest locked-block tests are vacuous: fixture blocks share one position
## What

In `src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListenerTest.kt`, `createBasicFixture()` builds the clicked block and its `UP`/`DOWN` neighbours with `testUtils.createMockBlock(world)`, which stubs `x`, `y` and `z` to `0` for all three. All three therefore map to the *same* `MfBlockPosition` (a data class over `worldId`, `x`, `y`, `z`), because they also share the same mock `World`.

`setupLockedBlock(...)` then does:

```kotlin
doReturn(lockedBlock).`when`(lockService).getLockedBlock(blockPosition)
// Mock the upward block position as well for bisected blocks
val upBlock = fixture.block.getRelative(BlockFace.UP)
val upBlockPosition = MfBlockPosition.fromBukkitBlock(upBlock)
doReturn(null).`when`(lockService).getLockedBlock(upBlockPosition)
```

`upBlockPosition == blockPosition`, so the second stub replaces the first and `getLockedBlock` returns `null` for every position. No test that calls `setupLockedBlock` ever reaches the locked-block branch of `PlayerInteractListener.applyProtections`.

## Why it matters

The three affected tests all assert `verifyEventNotCancelled()`, which is also what happens when no lock is found at all, so they pass for the wrong reason and would keep passing if lock protection were removed entirely:

- `onPlayerInteract_LockedBlock_NonOwnerWithBypassEnabled_ShouldAllowInteraction`
- `onPlayerInteract_LockedBlock_PlayerIsAccessor_ShouldAllowInteraction`
- (and any future test built on `setupLockedBlock`)

## Suggested fix

Give the neighbour blocks distinct coordinates in `createBasicFixture()`, e.g. `testUtils.createMockBlock(world, 0, 1, 0)` and `testUtils.createMockBlock(world, 0, -1, 0)`.

Note this is not a one-line change: once the stubs take effect the locked-block branch is actually executed, and it calls `plugin.server.scheduler.runTaskAsynchronously(...)` to look up the lock owner's name. `medievalFactions.server` is not stubbed in this test class, so the tests will NPE until a mocked `Server`/`BukkitScheduler` is added to the fixture. Both need doing together.

## How this was found

While adding physical-interaction tests in #1976. A test for the locked-block message path was dropped from that PR rather than built on the broken fixture.

### LINKED ISSUE #1978: Locked blocks still flood chat on physical interactions (pressure plate / farmland)
Follow-up to #1957 / PR #1976.

PR #1976 suppresses the per-tick protection message for `Action.PHYSICAL` at three sites in `PlayerInteractListener.applyProtections`: `CannotInteractBlockInWilderness`, `FactionTerritoryProtectionBypassed`, and `CannotInteractWithBlockInFactionTerritory`.

The **locked-block** branch, which runs *earlier* in the same function, was not covered:

- `src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:158` — `LockProtectionBypassed`
- `src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:166` — `BlockLocked`

`lock()` (same file, line 251) applies no block-type restriction, so a player can `/mf lock` a pressure plate, tripwire, or farmland block inside their own faction's claim. A non-accessor standing on it produces the exact symptom reported in #1957 — one message per tick — and it is *worse* than the faction-territory case, because each of those messages is emitted from a `runTaskAsynchronously` block that performs a `playerService.getPlayer(...)` lookup. That is one scheduled async task plus one player lookup **per tick, per player, per locked physical block**.

## Suggested fix

Reuse the `suppressProtectionMessages` flag PR #1976 already introduces at line 113, and skip *scheduling the task at all* (not just the `sendMessage`) so the per-tick player lookup goes away too. `event.isCancelled = true` at line 169 must stay outside the guard — the protection itself must not change.

## Why this wasn't done in #1976

It cannot be tested against the current fixture. Two blockers, both pre-existing:

1. **#1977** — `setupLockedBlock` stubs `getLockedBlock` for the block and then for `block.getRelative(UP)`, but `TestUtils.createMockBlock` gives every block the same `(0,0,0)` position in the same world, so the second stub overwrites the first with `null`. No existing locked-block test ever sees a locked block.
2. `PlayerInteractListenerTest` never mocks `medievalFactions.server`, so `plugin.server.scheduler.runTaskAsynchronously` would NPE the moment a test actually reached the locked-block branch. (This is masked today by blocker 1.)

So #1977 should be fixed first, or at minimum the scheduler mocked, before this fix lands with real regression coverage.

dmccoystephenson:
## Summary

- **#1978 — locked blocks still flood chat on physical interactions.** PR #1976 suppressed the per-tick protection message for `Action.PHYSICAL` in the wilderness, faction-territory and bypass-notice branches of `PlayerInteractListener.applyProtections`, but not in the **locked-block** branch that runs earlier in the same function. `lock()` applies no block-type restriction, so a player can `/mf lock` a pressure plate, tripwire or farmland block — and a non-accessor standing on it got one message *per tick*, each emitted from a `runTaskAsynchronously` block doing a `playerService.getPlayer(...)` owner lookup. This reuses the existing `suppressProtectionMessages` flag and skips **scheduling the task at all**, so the per-tick lookup goes away too. `event.isCancelled = true` stays outside the guard — the protection itself is unchanged.
- **#1977 — locked-block tests were vacuous.** `MfBlockPosition` is a data class over `(worldId, x, y, z)` and `TestUtils.createMockBlock` defaults every block to `(0, 0, 0)`, so the clicked block and its `UP`/`DOWN` neighbours mapped to the *same* position. `setupLockedBlock`'s second stub therefore overwrote the first with `null`, and no test ever reached the locked-block branch — the two existing locked-block tests passed for the wrong reason and would have kept passing if lock protection were deleted. Neighbours now get real coordinates and both faces are stubbed.
- As #1977 predicted, that fix alone makes the tests NPE, because `medievalFactions.server` was never mocked. Added a `mockScheduler()` following the existing `BlockBreakListenerTest` pattern, plus a `runScheduledAsyncTasks()` helper that **captures and runs** the dispatched `Runnable`s so the message logic is genuinely exercised rather than merely asserted to have been scheduled.

### Test coverage added

| Test | Asserts |
|---|---|
| `LockedBlock_NonOwnerWithoutBypass_ShouldCancelAndNotifyPlayer` | new — the locked-block branch is actually reached: cancelled, and the message is sent when the captured task runs |
| `PhysicalActionOnLockedBlock_ShouldCancelWithoutSchedulingOwnerLookup` | new — regression guard for #1978: still cancelled, **no** task scheduled, no message |
| `PhysicalActionOnLockedBlockWithBypassEnabled_ShouldNotScheduleOwnerLookup` | new — same for the bypass notice |
| `LockedBlock_NonOwnerWithBypassEnabled_ShouldAllowInteraction` | strengthened — now runs the captured task and asserts the player *is* notified on a deliberate right-click |
| `LockedBlock_PlayerIsAccessor_ShouldAllowInteraction` | strengthened — asserts no task is scheduled for an accessor |

## Test plan

- [x] `build` green in CI on head `8a690cf3` — 355 tests completed, 0 failed ([run 30215890703](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30215890703))
- [x] `docker-build` green on the same run
- [x] **Regression evidence (empirical, not reasoned):** the fix was temporarily reverted on this branch (`a86e1041`, since force-pushed away) and CI failed with *exactly* the two new `PhysicalAction…LockedBlock…` tests and nothing else ([run 30216194956](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30216194956), 355 tests, 2 failed). The branch was reset to `8a690cf3` and re-verified green.
- [x] Existing physical-interaction tests from #1976 (faction territory, wilderness, bypass) untouched and still asserting the same behavior

**Anchor note:** local `./gradlew` cannot run in this environment (Gradle 7.6.4 vs the only available JVM: `Unsupported class file major version 65`), and no container runtime is available. Per the loop's UNVERIFIED rule the anchor is the **CI check on this PR's head SHA**, not a local run. The changed files are Kotlin sources fully inside the Gradle build, so CI's scope does cover them.

## Data safety

No schema migrations; no DPC contract change; no `config.yml` / `plugin.yml` / `lang/` change. No new user-facing strings — the existing `BlockLocked`, `LockProtectionBypassed` and `UnknownPlayer` keys are reused unmodified. The only production change is a conditional around two `runTaskAsynchronously` calls; the `event.isCancelled = true` that enforces lock protection is deliberately left outside it. No modified path matches the do-not-auto-merge list.

## Reviewer note

Suppressing the *bypass* notice on physical interactions makes it silent for admins, not merely throttled. That mirrors what #1976 already did to the faction-territory bypass notice, but it is a behavior change worth an explicit nod. Full reasoning, and three other judgment calls, are in the self-review comment below.

## Filed, not fixed here

#1981 — the `mfPlayer == null` branch of the same function also schedules an async player save per tick for an unregistered player. Same per-tick shape, unrelated to locks, left out to keep this PR scoped.

## Deferred this cycle

The rest of the open backlog was not touched: this cycle was scoped to the #1977/#1978 pair, which have a dependency relationship (#1977 is the prerequisite for testing #1978) and were the freshest, best-specified issues. The remaining open issues are feature requests, unreproduced user bug reports, or (in the case of #1943 / #1914 / #1947) already have open `copilot/*` PRs.

Closes #1977
Closes #1978

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
+## [Unreleased]
+
+### Fixed
+- Standing on a *locked* pressure plate (or tripwire, or farmland) you are not allowed to use no longer floods chat with a lock message every tick. This covers the one path missed by the previous physical-interaction fix, and it also stops the per-tick scheduled task and block-owner lookup that each of those messages performed. The lock protection itself is unchanged — the interaction is still blocked.
+
 ## [6.0.0-SNAPSHOT-7-25-2026] – 2026-07-25
 
 ### Changed
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
@@ -151,21 +151,27 @@ class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
             if (event.player.uniqueId.toString() !in (lockedBlock.accessors + lockedBlock.playerId).map(MfPlayerId::value)) {
                 // Check if player has bypass permission from mf.bypass or faction permission
                 if ((mfPlayer.isBypassEnabled && event.player.hasPermission("mf.bypass")) || hasLockBypassPermission(mfPlayer)) {
-                    plugin.server.scheduler.runTaskAsynchronously(
-                        plugin,
-                        Runnable {
-                            val owner = playerService.getPlayer(lockedBlock.playerId)
-                            event.player.sendMessage("$RED${plugin.language["LockProtectionBypassed", owner?.toBukkit()?.name ?: plugin.language["UnknownPlayer"]]}")
-                        }
-                    )
+                    // The whole task is skipped when suppressing, not just the message: each one performs
+                    // an owner lookup, so scheduling it per tick is the expensive half of the problem.
+                    if (!suppressProtectionMessages) {
+                        plugin.server.scheduler.runTaskAsynchronously(
+                            plugin,
+                            Runnable {
+                                val owner = playerService.getPlayer(lockedBlock.playerId)
+                                event.player.sendMessage("$RED${plugin.language["LockProtectionBypassed", owner?.toBukkit()?.name ?: plugin.language["UnknownPlayer"]]}")
+                            }
+                        )
+                    }
                 } else {
-                    plugin.server.scheduler.runTaskAsynchronously(
-                        plugin,
-                        Runnable {
-                            val owner = playerService.getPlayer(lockedBlock.playerId)
-                            event.player.sendMessage("$RED${plugin.language["BlockLocked", owner?.toBukkit()?.name ?: plugin.language["UnknownPlayer"]]}")
-                        }
-                    )
+                    if (!suppressProtectionMessages) {
+                        plugin.server.scheduler.runTaskAsynchronously(
+                            plugin,
+                            Runnable {
+                                val owner = playerService.getPlayer(lockedBlock.playerId)
+                                event.player.sendMessage("$RED${plugin.language["BlockLocked", owner?.toBukkit()?.name ?: plugin.language["UnknownPlayer"]]}")
+                            }
+                        )
+                    }
                     event.isCancelled = true
                 }
                 return
--- a/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListenerTest.kt
+++ b/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListenerTest.kt
@@ -17,6 +17,7 @@ import com.dansplugins.factionsystem.player.MfPlayerId
 import com.dansplugins.factionsystem.player.MfPlayerService
 import com.dansplugins.factionsystem.relationship.MfFactionRelationshipService
 import org.bukkit.Material
+import org.bukkit.Server
 import org.bukkit.World
 import org.bukkit.block.Block
 import org.bukkit.block.BlockFace
@@ -29,10 +30,14 @@ import org.bukkit.event.block.Action
 import org.bukkit.event.player.PlayerInteractEvent
 import org.bukkit.inventory.EquipmentSlot
 import org.bukkit.inventory.ItemStack
+import org.bukkit.scheduler.BukkitScheduler
 import org.junit.jupiter.api.BeforeEach
 import org.junit.jupiter.api.Test
+import org.mockito.ArgumentCaptor
 import org.mockito.ArgumentMatchers.any
 import org.mockito.ArgumentMatchers.anyString
+import org.mockito.ArgumentMatchers.eq
+import org.mockito.Mockito.atLeast
 import org.mockito.Mockito.doReturn
 import org.mockito.Mockito.mock
 import org.mockito.Mockito.never
@@ -84,6 +89,7 @@ class PlayerInteractListenerTest {
     private lateinit var lockService: MfLockService
     private lateinit var relationshipService: MfFactionRelationshipService
     private lateinit var factionService: com.dansplugins.factionsystem.faction.MfFactionService
+    private lateinit var scheduler: BukkitScheduler
     private lateinit var uut: PlayerInteractListener
 
     // Common test constants
@@ -96,6 +102,7 @@ class PlayerInteractListenerTest {
         medievalFactions = mock(MedievalFactions::class.java)
         mockServices()
         mockLanguageSystem()
+        mockScheduler()
         uut = PlayerInteractListener(medievalFactions)
     }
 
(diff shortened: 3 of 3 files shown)


dmccoystephenson:
Self-review rubric (anchored on CI run [30215890703](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30215890703) on head `8a690cf3`):

- **CI:** PASS — `build` and `docker-build` both green on `8a690cf3`, 355 tests completed, 0 failed. Local `./gradlew` cannot run in this environment (Gradle 7.6.4 vs JVM 21: `Unsupported class file major version 65`) and no container runtime is available, so per the loop's UNVERIFIED rule the CI check on the exact head SHA is the anchor. CI's scope does cover every changed file.
- **Tests-fix (empirical, not reasoned):** PASS — the fix was temporarily reverted on this branch (commit `a86e1041`, since force-pushed away) and CI run [30216194956](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30216194956) failed with **exactly** the two new regression tests and nothing else:
  ```
  PlayerInteractListenerTest > onPlayerInteract_PhysicalActionOnLockedBlockWithBypassEnabled_ShouldNotScheduleOwnerLookup() FAILED
  PlayerInteractListenerTest > onPlayerInteract_PhysicalActionOnLockedBlock_ShouldCancelWithoutSchedulingOwnerLookup() FAILED
  355 tests completed, 2 failed
  ```
  The branch was then reset back to `8a690cf3` and re-verified green. The discriminator is the `verify(scheduler, never())` assertion, not a message assertion — without the guard the per-tick task is still scheduled even though the message never reaches the player in-test, so this is a real signal rather than a false negative.
- **Scope:** PASS — 3 files, +139/-24. `PlayerInteractListener.kt` is the fix, `PlayerInteractListenerTest.kt` is the fixture repair plus coverage, `CHANGELOG.md` is the required doc entry. No unrelated formatting, renames or comment churn; no tree-wide reformat.
- **Tests-new:** PASS — no new public methods are introduced, so this reduces to covering the changed path: 3 new tests plus 2 strengthened existing ones, all exercising the locked-block branch that was previously unreachable.
- **Migrations-additive:** PASS — `git diff origin/main...HEAD -- '*db/migration/*'` is empty. No schema change of any kind.
- **Lang-externalized:** PASS — no string literal added to any `.kt` file. The existing `BlockLocked`, `LockProtectionBypassed` and `UnknownPlayer` keys are reused unmodified; no `lang/*.properties` file is touched, so the Latin-1 encoding hazard does not arise.
- **DPC-contract:** N/A — nothing under `dpc/` or `pacts/` is in the diff.
- **Main-thread-safety:** PASS — the change only ever *removes* an async dispatch; it adds none, and touches no Bukkit access outside the existing main-thread event handler.
- **Plugin.yml / Config-defaults:** N/A — neither file is in the diff; no new command, permission or config key.
- **Sibling structure:** PASS — `mockScheduler()` is copied from the existing `BlockBreakListenerTest` helper of the same name, and the new tests sit in the file's existing `// --- Physical interaction tests ---` section following the comment-and-Arrange/Act/Assert style of the #1976 tests immediately above them.
- **Sibling renames:** N/A — nothing renamed.
- **Docs:** PASS — `CHANGELOG.md` has an `[Unreleased] / Fixed` entry. Swept the other sources of truth: `CONFIG.md:126` already documents the wilderness-alert suppression and no new config key is added; `COMMANDS.md` and `FACTION_FLAGS.md` are unaffected; `USER_GUIDE.md`'s "Locking Important Blocks" section describes how to lock a block, not the per-interaction notice, so it needs no change; `DATABASE_QUERYING.md` / `DEPENDENCY_USAGE.md` / both READMEs unaffected.
- **Issue resolution:** PASS — #1977's named surface (`createBasicFixture`, `setupLockedBlock`, the unmocked `medievalFactions.server`) is changed, and the locked-block branch is now demonstrably reachable. #1978's two named call sites (`PlayerInteractListener.kt` `LockProtectionBypassed` and `BlockLocked`) are both guarded. Neither is partially resolved.

**Judgment calls a reviewer should weigh, rather than an all-PASS claim:**

1. **The bypass notice is now silent for physical interactions, not just throttled.** An admin with `mf.bypass` who walks over a locked pressure plate gets *no* message at all now. That is deliberate and mirrors exactly what #1976 already did to the faction-territory bypass notice, but it is a real behavior change for admins, not purely a spam fix, and it is worth a maintainer's explicit agreement.
2. **Two existing tests were passing for the wrong reason and are only now meaningful.** `LockedBlock_NonOwnerWithBypassEnabled_ShouldAllowInteraction` and `LockedBlock_PlayerIsAccessor_ShouldAllowInteraction` never reached the locked-block branch before this PR (that is #1977), so their green history says nothing about lock protection. They have been strengthened rather than left as-is, but reviewers should not read their prior passing as prior coverage.
3. **`runScheduledAsyncTasks()` runs every captured task, not a specific one.** In these tests only one task is ever dispatched, so it is unambiguous today. A future test that reaches two different scheduling sites would need a more targeted assertion.
4. **Out of diff, noted for follow-up:** the `mfPlayer == null` branch (`PlayerInteractListener.kt:117`) also schedules an async `playerService.save(...)` on *every* interaction including `PHYSICAL`, and the save is asynchronous, so an unregistered player standing on any protected block can queue a duplicate save per tick until the first one lands. That is pre-existing and unrelated to lock protection, so it is deliberately not fixed here.

**Data safety:** no schema migration, no DPC wire change, no `config.yml` / `plugin.yml` / `lang/` change. No modified path matches the do-not-auto-merge list.

---

PR #1982, opened 2026-07-27
Dedupe async player-save on repeated physical interactions

### LINKED ISSUE #1981: Unregistered player on a protected physical block queues a duplicate async save per tick
Found while fixing #1978; deliberately left out of PR #1980 to keep that change scoped to the locked-block message path.

`PlayerInteractListener.applyProtections` handles a player with no `MfPlayer` record by cancelling the event and scheduling an async save:

```kotlin
if (mfPlayer == null) {
    event.isCancelled = true
    plugin.server.scheduler.runTaskAsynchronously(
        plugin,
        Runnable {
            playerService.save(MfPlayer(plugin, event.player)).onFailure { ... }
        }
    )
    return
}
```

`src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:115-128`

For `Action.RIGHT_CLICK_BLOCK` this fires once per deliberate click, which is fine. For `Action.PHYSICAL` it fires **once per tick** for as long as the player stands on the block (this is the same per-tick property behind #1957 and #1978). The save is asynchronous, so every tick between the first dispatch and the point where `playerService.getPlayer(...)` starts returning the saved record queues another `save` — one scheduled task and one write attempt each.

## Severity

Low. The window is bounded by how long the first async save takes to complete and populate the service's cache, so it is a handful of redundant writes for a brand-new player, not an unbounded loop. It is worth fixing mainly because the "one scheduled task per tick" shape is exactly what #1957 and #1978 were about, and this is the last instance of it in this listener.

## Suggested fix

Track in-flight saves per player id (or check the cache inside the scheduled task before writing) so only the first physical interaction dispatches a save. Note that `suppressProtectionMessages` is **not** the right guard here: skipping the save entirely on `PHYSICAL` would mean a player who only ever steps on a pressure plate never gets a record created.

Regression coverage is now possible: `PlayerInteractListenerTest` mocks `medievalFactions.server` as of PR #1980, and `runScheduledAsyncTasks()` captures and runs dispatched `Runnable`s.

dmccoystephenson:
## Summary
- An unregistered player standing on a protected physical block (pressure plate, tripwire, farmland) fires `Action.PHYSICAL` once per tick, and `PlayerInteractListener.applyProtections` previously queued a fresh async `MfPlayer` save for every one of those ticks until the first save completed and populated `MfPlayerService`'s cache.
- Added a per-listener set (`playersWithPendingSave`) tracking player ids with an in-flight save; only the first physical interaction for a given player dispatches a save, and the id is released (success or failure) when that save finishes, so a later interaction can dispatch again if the record still isn't visible.
- `event.isCancelled = true` still runs unconditionally — the protection itself is unchanged, only the redundant scheduling/write is deduped.

## Test plan
- [x] Added `onPlayerInteract_PhysicalActionByUnregisteredPlayer_ShouldOnlyScheduleOneSavePerTick` — two physical interactions before the save runs only schedule one async task.
- [x] Added `onPlayerInteract_PhysicalActionByUnregisteredPlayer_ShouldScheduleAnotherSaveOnceThePriorOneCompletes` — after the first scheduled save runs (success), a subsequent interaction schedules a second save.
- [x] `./gradlew compileKotlin` — passes locally.
- [ ] `./gradlew clean test` — **UNVERIFIED locally**: `compileJava` dies in this sandbox on a pre-existing Gradle user-home compile-cache lock (`Unexpected lock protocol found in lock file. Expected 3, found 0`), unrelated to this change and present before any edits. Relying on the CI `build` check on this PR's head SHA as the verification anchor per this repo's dev-loop UNVERIFIED-anchor rule.

## Data-safety statement
No schema migrations; no DPC contract change; no `config.yml` default change; no `lang/` changes. Purely an in-memory dedup guard around an existing async save path.

Closes #1981

---

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ### Fixed
 - Standing on a *locked* pressure plate (or tripwire, or farmland) you are not allowed to use no longer floods chat with a lock message every tick. This covers the one path missed by the previous physical-interaction fix, and it also stops the per-tick scheduled task and block-owner lookup that each of those messages performed. The lock protection itself is unchanged — the interaction is still blocked.
+- A player with no `MfPlayer` record yet who stands on a protected physical block (pressure plate, tripwire, farmland) no longer queues a duplicate async player-save every tick. Only the first physical interaction dispatches a save; further ticks are ignored until that save completes.
 
 ## [6.0.0-SNAPSHOT-7-25-2026] – 2026-07-25
 
--- a/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
+++ b/src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt
@@ -37,10 +37,16 @@ import org.bukkit.event.Listener
 import org.bukkit.event.block.Action.PHYSICAL
 import org.bukkit.event.player.PlayerInteractEvent
 import org.bukkit.inventory.EquipmentSlot.HAND
+import java.util.concurrent.ConcurrentHashMap
 import java.util.logging.Level.SEVERE
 import org.bukkit.block.data.type.Gate as FenceGateData
 
 class PlayerInteractListener(private val plugin: MedievalFactions) : Listener {
+
+    // Tracks players whose MfPlayer record is currently being created asynchronously, so that
+    // Action.PHYSICAL (which fires once per tick while the player stands on the block) doesn't
+    // queue a duplicate save for every tick until the first one completes.
+    private val playersWithPendingSave: MutableSet<MfPlayerId> = ConcurrentHashMap.newKeySet()
 
     @EventHandler
     fun onPlayerInteract(event: PlayerInteractEvent) {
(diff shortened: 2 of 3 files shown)


dmccoystephenson:
Self-review rubric (local verification recovered this pass — found that `-Dgradle.user.home=<repo>/.gradle-local` sidesteps the sandbox's corrupted default `~/.gradle` compile-cache lock, so this could run for real rather than relying solely on CI):

- **Scope:** PASS — every changed file is necessary: the listener fix, its test file, and the one CHANGELOG entry. No unrelated churn.
- **Tests-new:** PASS — two new tests cover the dedup guard: one save in flight is not duplicated across repeated ticks, and a completed save releases the guard so a later tick can dispatch again.
- **Tests-fix (empirical):** PASS — `git checkout <merge-base> -- PlayerInteractListener.kt` (reverting only the production fix, keeping the new tests) then `./gradlew test --tests ...PlayerInteractListenerTest`: `onPlayerInteract_PhysicalActionByUnregisteredPlayer_ShouldOnlyScheduleOneSavePerTick` FAILED (1/39) as expected. Restored the fix (`git checkout HEAD --`), reran `./gradlew clean test`: full suite green. Direct fail→pass evidence.
- **Sibling structure:** PASS — new tests sit beside the existing `Action.PHYSICAL` regression tests for #1978 in the same file/class, same Arrange/Act/Assert shape. The `anyMfPlayer()` helper mirrors the `anyFaction()` generic-indirection workaround already established in `MfFactionAddMemberCommandTest` for the identical Mockito/Kotlin `any()`-null-check issue (hit and fixed during this PR — see commit `ba2f634d`).
- **Sibling renames:** N/A — no renames in this diff.
- **Docs:** PASS — `CHANGELOG.md` `[Unreleased] > Fixed` has a new entry; no command, config, or faction-flag surface changed, so `COMMANDS.md`/`CONFIG.md`/`FACTION_FLAGS.md`/`USER_GUIDE.md` are unaffected.
- **Issue resolution:** PASS — `Closes #1981` and the diff changes exactly the code path `#1981` names (`PlayerInteractListener.applyProtections`, the `mfPlayer == null` branch).
- **CI:** PASS — `build` and `docker-build` both green on head `739bac8f` (https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30248090076).

Repo-specific:
- **Migrations-additive:** PASS (vacuous) — no migration files touched.
- **Lang-externalized:** PASS (vacuous) — no new user-facing string; the existing `BlockInteractFailedToSavePlayer` lang key is unchanged, only its dispatch is deduped.
- **DPC-contract:** PASS (vacuous) — `dpc/` untouched.
- **Main-thread-safety:** PASS — the guard set (`playersWithPendingSave`, a `ConcurrentHashMap`-backed set) is mutated both on the calling thread (`add`) and inside the async runnable (`remove` in `finally`), which is exactly why it needs to be a concurrent set rather than a plain `HashSet`; no new Bukkit API call was moved off the main thread.
- **Plugin.yml:** PASS (vacuous) — no command/permission changes.
- **Config-defaults:** PASS (vacuous) — no `config.yml` changes.

**Data-safety statement:** No schema migrations; no DPC contract change; no `config.yml` default change; no `lang/` changes. In-memory dedup guard only — no persisted data shape changes. Does not match any do-not-auto-merge path (migrations, `dpc/`, `config.yml`, `plugin.yml`, `lang/*` removals, `build.gradle`/workflows, `security/`).

All rubric items pass; no outstanding judgment calls to flag for a human reviewer.

---

Issue #1988, opened 2026-07-31
mf.force.kick and mf.force.rename do not bypass the target faction's role check, unlike mf.force.flag


dmccoystephenson:
Found while documenting the `mf.force.*` overrides for #1986. Filed rather than fixed, because a documentation cycle should not silently change behavior — and because which way this should go is a design call.

## Finding

`mf.force.kick` and `mf.force.rename` let an admin *name* a faction other than their own, but the target faction's own role permission check still runs afterwards against the sender. Since an admin is normally not a member of the faction they are acting on, `getRole()` returns `null` and the command is refused — so the force permission does not actually let them force anything.

**`MfFactionKickCommand.kt`** — the force permission only selects which faction is resolved:

```kotlin
val faction = if (args.size > 1 && hasForcePermission) {
    factionService.getFaction(args.dropLast(1).joinToString(" "))
} else {
    factionService.getFaction(mfPlayer.id)
}
...
val role = faction.getRole(mfPlayer.id)
if (role == null || !role.hasPermission(faction, plugin.factionPermissions.kick)) {
    sender.sendMessage("$RED${plugin.language["CommandFactionKickNoFactionPermission"]}")
    return@Runnable
}
```

**`MfFactionSetNameCommand.kt`** — same shape:

```kotlin
if (hasForcePermission) {
    val unquotedArgs = args.unquote()
    if (unquotedArgs.size > 1) {
        faction = factionService.getFaction(MfFactionId(unquotedArgs[0])) ?: factionService.getFaction(unquotedArgs[0])
        name = unquotedArgs.drop(1).joinToString(" ")
    }
}
...
val role = faction.getRole(mfPlayer.id)
if (role == null || !role.hasPermission(faction, plugin.factionPermissions.changeName)) {
    player.sendMessage("$RED${plugin.language["CommandFactionSetNameNoFactionPermission"]}")
    return@Runnable
}
```

Net effect: an op with `mf.force.kick` who is not in `SomeFaction` runs `/f kick SomeFaction PlayerName` and gets "you do not have permission" from the faction role system.

## Contrast: `mf.force.flag` does bypass the role check

`MfFactionFlagListCommand.kt` / `MfFactionFlagSetCommand.kt` branch on whether a target faction was given, and skip the role check entirely when one was:

```kotlin
return if (targetFaction != null) {
    if (!sender.hasPermission("mf.force.flag")) { ...deny...; return false }
} else {
[...]

Issue #1987, opened 2026-07-31
mf.force.claim and mf.force.unclaim are declared in plugin.yml but never read


dmccoystephenson:
Split out from #1986 so the documentation half of that issue can be closed independently.

## Finding

`src/main/resources/plugin.yml` declares two permission nodes:

```yaml
  mf.force.claim:
    description: Allows claiming land for other factions
    default: op
  mf.force.unclaim:
    description: Allows unclaiming land for other factions
    default: op
```

Both are also listed as children of `mf.force.*`, which is itself granted by `mf.admin`. But a full source search finds zero references to either node anywhere in the plugin:

```
$ grep -rn "force.claim\|force.unclaim" src/main/kotlin
(no matches)
```

The claim and unclaim commands never check for them, and neither does any listener. Granting either node currently has no effect at all.

Every other `mf.force.*` node is read somewhere in the source (`addmember`, `bonuspower`, `flag`, `join`, `kick`, `power`, `rename`, `unlock`), so these two are the odd ones out.

## Decision needed

Either:

1. **Implement the feature** the descriptions promise — an admin form of `/f claim` / `/f unclaim` that takes a faction name and claims/unclaims on that faction's behalf; or
2. **Remove the dead declarations** from `plugin.yml` (both the node definitions and the two `mf.force.*` children entries).

This is a maintainer call about original intent, not a mechanical fix, so it is filed rather than guessed at.

## Interim

The PR for #1986 documents in `COMMANDS.md` that these two nodes currently do nothing, so admins do not grant them expecting a behavior that is not there. That note should be removed or updated by whichever option above is taken.

PR #1985, opened 2026-07-30
docs: document admin view/disband-other permission gates

dmccoystephenson:
## Summary
- Documents three admin-facing command behaviors in COMMANDS.md that exist in the shipped code but had no documentation at all:
  - `/faction info [player|faction]` — with `mf.info.other` (default: `true`), an argument targets another player's or named faction's info instead of the sender's own (also lifts the "must be a player" requirement, allowing console usage). See `MfFactionInfoCommand.kt`.
  - `/faction power [player|faction]` — with `mf.power.view.other` (default: `true`), same pattern for viewing another player's/faction's power. See `MfFactionPowerCommand.kt`.
  - `/faction disband [faction]` — with `mf.disband.others` (default: `op`), a faction name argument disbands any faction regardless of member count, bypassing the normal "must be down to one member" self-disband guard. See `MfFactionDisbandCommand.kt`.
- Each finding was verified directly against the command implementation and cross-checked against the `default:` value declared in `plugin.yml` before writing the note (two of the three default to `true`, not `op`, which is itself worth flagging since it means any player can already view another faction's info/power today).

## Test plan
- [x] Docs-only change — no source touched, nothing to compile or test.
- [x] Verified `mf.info.other` / `mf.power.view.other` / `mf.disband.others` permission checks and default values by reading `MfFactionInfoCommand.kt`, `MfFactionPowerCommand.kt`, `MfFactionDisbandCommand.kt`, and `plugin.yml`.

## Data-safety statement
No schema migrations; no DPC contract change; no `config.yml` default change; no `plugin.yml` change (docs only, describing existing registered permissions).

---

_This PR description was drafted during a Gardener session ([Stephenson-Software/gardener](https://github.com/Stephenson-Software/gardener))._

PR #1989, opened 2026-07-31
docs: document the mf.force.* permission overrides in COMMANDS.md

### LINKED ISSUE #1986: Document mf.force.* command overrides in COMMANDS.md; mf.force.claim/unclaim appear unused
## Summary

Follow-up from a documentation sweep in PR #1985. Several `mf.force.*` permission nodes are declared in `plugin.yml` and change the behavior of existing commands, but that behavior is not documented anywhere in `COMMANDS.md`. Two other `mf.force.*` nodes are declared and grouped under `mf.admin` but are never referenced anywhere in the Kotlin source at all.

All of the following was verified by reading the command source and `plugin.yml`, not assumed.

### Implemented but undocumented force overrides

- **`mf.force.kick`** (`MfFactionKickCommand.kt`) — with this permission, `/f kick [faction] [player]` (two args) kicks a player from a faction other than the sender's own, bypassing the normal role/permission check. Currently `COMMANDS.md`'s `/faction kick [player]` entry only documents the single-arg, own-faction case.
- **`mf.force.join`** (`MfFactionJoinCommand.kt`) — lets a player join a faction without having been invited. Not mentioned under `/faction join [faction]`.
- **`mf.force.rename`** (`MfFactionSetNameCommand.kt`) — lets a player rename a faction that isn't their own / bypasses the normal rename permission check. Not mentioned under `/faction set [name|description|prefix] [value]`.
- **`mf.force.flag`** (`MfFactionFlagListCommand.kt`, `MfFactionFlagSetCommand.kt`) — bypasses the normal per-faction flag permission checks for both listing and setting flags on another faction. Not mentioned under `/faction flag [list|set]`.
- **`mf.force.unlock`** (`PlayerInteractListener.kt`, `BlockBreakListener.kt`) — bypasses lock protection so a player can interact with or break another faction's locked blocks. This one isn't a command at all (it's a listener bypass), so it doesn't fit neatly into a command doc entry — worth a short mention in `COMMANDS.md`'s Locks & Access Control section or the Permission Groups section regardless.

`mf.force.power`, `mf.force.bonuspower`, and `mf.force.addmember` (the last from PR #1919, not yet merged) already have per-command documentation in `COMMANDS.md` and can be used as the template for wording/format.

### Declared but unused permission nodes (not a docs gap — a code gap)

- **`mf.force.claim`** and **`mf.force.unclaim`** are declared in `plugin.yml` (`Allows claiming/unclaiming land for other factions`) and listed as children of `mf.admin`, but a full source search (`grep -rn "force.claim\|force.unclaim" src/main/kotlin`) finds zero references anywhere in the Kotlin code. Neither `MfFactionClaimCommand.kt`/`MfFactionClaimCircleCommand.kt` nor the unclaim command check for these permissions. Either the "claim/unclaim land for another faction" admin feature was never implemented, or these nodes are vestigial and should be removed from `plugin.yml`. This needs a maintainer decision on intent before either the feature is built or the dead permission nodes are cleaned up — not something to guess at silently.

### Suggested scope for a follow-up PR

1. Add per-command "Notes" lines to `COMMANDS.md` for `mf.force.kick`, `mf.force.join`, `mf.force.rename`, `mf.force.flag`, and `mf.force.unlock`, mirroring the existing `mf.force.power`/`mf.force.bonuspower` style.
2. Raise the `mf.force.claim`/`mf.force.unclaim` dead-node finding with the maintainer separately (implement the feature, or remove the unused permission declarations from `plugin.yml`) — this is a decision, not a mechanical fix, so it's flagged here rather than auto-resolved.

Not fixed in this cycle to keep PR #1985 scoped to the three already-verified, standalone (`mf.info.other` / `mf.power.view.other` / `mf.disband.others`) gaps.

---

dmccoystephenson:
## Summary

Documentation-only. `COMMANDS.md` never mentioned five of the ten `mf.force.*` permission nodes, even though each one changes the behaviour of a command an admin is likely to reach for. Every note below was written by reading the command/listener source, not the `plugin.yml` description.

- **`mf.force.join`** → `/f join` — join a faction you were never invited to. Without `-f` this shows a clickable confirmation prompt instead of the "not invited" error; `/f join FactionName -f` skips the prompt.
- **`mf.force.kick`** → `/f kick` — the two-argument `/f kick FactionName PlayerName` form that targets a faction other than your own.
- **`mf.force.rename`** → `/f set name` — the `/f set name FactionName NewName` form.
- **`mf.force.flag`** → `/f flag list|set` — a faction name/ID as the first argument, plus faction-name tab completion.
- **`mf.force.unlock`** → `/unlock` and block breaking — unlocking a block locked by *another player*.

Also in this PR:

- A **Force Permissions** index table under `Permission Groups`, since these nodes add no commands of their own and are therefore hard to find. It records that `mf.force.claim` / `mf.force.unclaim` are declared in `plugin.yml` but read nowhere in the source.
- `/f addmember`'s permission line corrected to `mf.force.addmember` **or** `mf.force.join` — `MfFactionAddMemberCommand.kt:21` accepts either.
- Three `##` sections that were missing from the table of contents (`Moderator Approval`, `DPC Community API`, `Permission Groups`), and the two `mf.approve` / `mf.dpc` bullets kept with the Admin Permissions list they belong to.

## Two findings filed rather than fixed

A documentation cycle should not silently change behaviour, so both of these are documented as they currently are and tracked separately:

- **#1987** — `mf.force.claim` and `mf.force.unclaim` are declared and granted by `mf.force.*` but never read anywhere in `src/main/kotlin`. Needs a maintainer call: implement the feature, or delete the dead declarations.
- **#1988** — `mf.force.kick` and `mf.force.rename` only widen *which faction* the command targets; the target faction's role check still runs against the sender afterwards, so an admin who is not a member of that faction is still refused. `mf.force.flag` genuinely bypasses the role check. The two behaviours disagree, and `plugin.yml`'s descriptions match the flag one. Fixing it would widen an op-default permission, so it wants a deliberate decision.

The `COMMANDS.md` notes describe today's actual behaviour, caveat included, and both notes point at the tracking issue so they get updated when a decision lands.

## Test plan

- [x] Every claim verified against source: `MfFactionJoinCommand.kt`, `MfFactionKickCommand.kt`, `MfFactionSetNameCommand.kt`, `MfFactionFlagListCommand.kt`, `MfFactionFlagSetCommand.kt`, `MfFactionAddMemberCommand.kt`, `PlayerInteractListener.kt`, `BlockBreakListener.kt`, `plugin.yml`.
- [x] Anchor links in the new table checked against the actual `###` headings in `COMMANDS.md`.
- [ ] CI `build` / `docker-build` green.

**Local build anchor:** UNVERIFIED-not-applicable. No Kotlin, resource, or build file is touched, so `./gradlew test` has nothing to say about this diff; CI on the head SHA is the anchor.

**No `CHANGELOG.md` entry:** no plugin behaviour changed. This follows the precedent of the previous docs-only commit on `main` (`aabad183`, PR #1985), which also added none.

## Data-safety statement

No schema migrations; no DPC contract change; no `config.yml` default change; no `plugin.yml` change; no lang-file change. Single-file documentation diff (`COMMANDS.md`, +32/-3).

## Deferred backlog

This cycle was scoped to #1986; the rest of the open issue backlog and the four remaining open PRs (#1915, #1918, #1944, #1948 — all currently conflicting with `main`) were not touched. #1919 was already groomed and self-reviewed in a previous cycle and is awaiting human approval.

Closes #1986

---

### DIFF
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -123,7 +123,7 @@ The main faction command can be accessed using any of the following aliases:
 
 **Example:** `/f flag set color #FF0000`
 
-**Notes:** With `mf.force.flag` (default: op), a faction name or ID can be given as the first argument to list or set flags on a faction other than your own (`/f flag list FactionName [page]`, `/f flag set FactionName [flag] [value]`). Unlike `mf.force.kick` and `mf.force.rename`, this genuinely bypasses the target faction's role permissions; setting a flag this way still additionally requires `mf.flag.set`. The permission also enables faction-name tab completion for both subcommands.
+**Notes:** With `mf.force.flag` (default: op), a faction name or ID can be given as the first argument to list or set flags on a faction other than your own (`/f flag list FactionName [page]`, `/f flag set FactionName [flag] [value]`). Unlike `mf.force.kick` and `mf.force.rename`, this genuinely bypasses the target faction's role permissions; the base `mf.flag.list` / `mf.flag.set` node is still required. The permission also enables faction-name tab completion for both subcommands.
 
 See [FACTION_FLAGS.md](FACTION_FLAGS.md) for a complete list of available flags.
 
@@ -365,7 +365,7 @@ See [FACTION_FLAGS.md](FACTION_FLAGS.md) for a complete list of available flags.
 - `/unlock` - Enables unlock mode
 - `/unlock cancel` - Cancels unlock mode
 
-**Notes:** Normally you can only unlock blocks you locked yourself. With `mf.force.unlock` (default: op), unlock mode also works on blocks locked by another player, and breaking another player's locked block unlocks it instead of being refused. A message is shown naming the lock owner whose protection was bypassed. (Faction roles can also grant a lock bypass independently of this permission.)
+**Notes:** Normally you can only unlock blocks you locked yourself. With `mf.force.unlock` (default: op), unlock mode also works on blocks locked by another player — a message naming the lock owner whose protection was bypassed is shown — and breaking another player's locked block unlocks it instead of being refused. (A faction role can also grant a lock bypass for unlock mode, independently of this permission.)
 
 ### `/accessors list`
 **Permission:** `mf.accessors.list` (default: true)  
--- a/COMMANDS.md
+++ b/COMMANDS.md
@@ -17,7 +17,10 @@ This document provides a comprehensive list of all commands available in the Med
 - [Power System](#power-system)
 - [Chat System](#chat-system)
 - [Applications](#applications)
+- [Moderator Approval](#moderator-approval)
 - [Admin Commands](#admin-commands)
+- [DPC Community API](#dpc-community-api)
+- [Permission Groups](#permission-groups)
 
 ## Command Aliases
 
@@ -75,7 +78,8 @@ The main faction command can be accessed using any of the following aliases:
 ### `/faction join [faction]` or `/f join [faction]`
 **Permission:** `mf.join` (default: true)  
 **Description:** Joins a faction that you have been invited to.  
-**Usage:** `/f join FactionName`
+**Usage:** `/f join FactionName`  
+**Notes:** With `mf.force.join` (default: op), you can join a faction you were never invited to. Running `/f join FactionName` without an invite shows a clickable confirmation prompt instead of the "not invited" error; appending `-f` (`/f join FactionName -f`) skips the prompt and joins immediately.
 
 ### `/faction leave` or `/f leave`
 **Permission:** `mf.leave` (default: true)  
@@ -90,7 +94,8 @@ The main faction command can be accessed using any of the following aliases:
 ### `/faction kick [player]` or `/f kick [player]`
 **Permission:** `mf.kick` (default: true)  
 **Description:** Kicks a player from your faction.  
-**Usage:** `/f kick PlayerName`
+**Usage:** `/f kick PlayerName`  
+**Notes:** With `mf.force.kick` (default: op), a faction name can be given before the player (`/f kick FactionName PlayerName`) to target a faction other than the one you are in. The faction-role check is still applied against the target faction, so you must also hold that faction's kick role permission — the permission selects which faction is acted on, it does not by itself override the faction's own permissions.
 
 ### `/faction set [name|description|prefix] [value]` or `/f set [name|description|prefix] [value]`
 **Permissions:** 
@@ -104,6 +109,8 @@ The main faction command can be accessed using any of the following aliases:
 - `/f set description "Our faction description"` - Sets faction description
 - `/f set prefix [TAG]` - Sets faction prefix
 
+**Notes:** With `mf.force.rename` (default: op), a faction name or ID can be given before the new name (`/f set name FactionName NewName`) to rename a faction other than the one you are in. As with `mf.force.kick`, the faction-role check still runs against the target faction, so you must also hold that faction's change-name role permission.
+
 ### `/faction flag [list|set]` or `/f flag [list|set]`
 **Permissions:** 
 - `mf.flag.list` (default: true)
@@ -115,6 +122,8 @@ The main faction command can be accessed using any of the following aliases:
 - `/f flag set [flag] [value]` - Sets a flag value
 
 **Example:** `/f flag set color #FF0000`
+
+**Notes:** With `mf.force.flag` (default: op), a faction name or ID can be given as the first argument to list or set flags on a faction other than your own (`/f flag list FactionName [page]`, `/f flag set FactionName [flag] [value]`). Unlike `mf.force.kick` and `mf.force.rename`, this genuinely bypasses the target faction's role permissions; setting a flag this way still additionally requires `mf.flag.set`. The permission also enables faction-name tab completion for both subcommands.
 
 See [FACTION_FLAGS.md](FACTION_FLAGS.md) for a complete list of available flags.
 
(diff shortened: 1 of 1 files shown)


dmccoystephenson:
## Self-review rubric

Anchored on CI green at head `97a7c8c4` (`build` pass 2m57s, `docker-build` pass 7m5s). Scored adversarially against the diff, not from memory.

- **Scope: PASS with a flag** — one file, `COMMANDS.md`, +32/-3 against `main`. Two of the changes are adjacent to #1986 rather than strictly inside it: the three `##` sections added to the table of contents (`Moderator Approval`, `DPC Community API`, `Permission Groups`), and moving the `mf.approve` / `mf.dpc` bullets so they stay attached to the Admin Permissions list above the new subsection. Both are pre-existing drift encountered while placing the new content, not unrelated churn — but they are drift that was chosen to be fixed here rather than filed, and a reviewer who wants this PR strictly scoped should say so.
- **Tests-new: N/A** — no production code in the diff; no new public method exists to test.
- **Tests-fix: N/A** — no bug is fixed here. The two behavioural defects found during the sweep (#1987, #1988) were deliberately left unfixed, so there is nothing to stash-and-prove.
- **Sibling structure: PASS with a note** — the per-command additions all use the existing `**Notes:**` line convention already used by `/f disband`, `/f create` and `/f power`. The Force Permissions index is the **first table in `COMMANDS.md`** (`grep -c '^|'` on `main` returns 0). Tables are established elsewhere in the repo's docs (`DATABASE_QUERYING.md`, both READMEs), and an 8-row permission→command→effect mapping reads badly as prose, so it was kept — but it is a deviation from this file's own prior style and worth a second opinion.
- **Sibling renames: N/A** — nothing renamed.
- **Docs (Phase 7 table): PASS** — `COMMANDS.md` is the only source of truth affected. No config key, faction flag, command registration, dependency, schema or user-facing behaviour changed, so `CONFIG.md`, `FACTION_FLAGS.md`, `USER_GUIDE.md`, `DATABASE_QUERYING.md`, `DEPENDENCY_USAGE.md` and both READMEs are unaffected. No `CHANGELOG.md` entry, matching the precedent of the previous docs-only commit on `main` (`aabad183`, PR #1985).
- **Issue resolution: PASS with a caveat** — #1986 asked for two things. Its documentation half (suggested scope item 1) is fully done: all five undocumented nodes are covered. Its code half (suggested scope item 2, the dead `mf.force.claim`/`mf.force.unclaim` declarations) is explicitly *not* resolved here — #1986 itself said to raise that separately, so it is carried into **#1987** with the full evidence rather than dropped. If you would rather #1986 stay open until the dead nodes are dealt with, remove the `Closes #1986` line before merging.
- **CI: PASS** — green on the exact head SHA.
- **Migrations-additive: PASS (vacuous)** — `git diff --name-only origin/main...HEAD` returns `COMMANDS.md` only; nothing under `db/migration/`.
- **Lang-externalized: PASS (vacuous)** — no `.kt`/`.java` in the diff, no new strings.
- **DPC-contract: PASS (vacuous)** — nothing under `dpc/` or `pacts/`.
- **Main-thread-safety: PASS (vacuous)** — no code.
- **Plugin.yml: PASS (vacuous)** — not modified. Note the inverse direction *was* checked: all ten `mf.force.*` nodes declared in `plugin.yml` are now accounted for in the docs, eight as working overrides and two as dead (#1987).
- **Config-defaults: PASS (vacuous)** — `config.yml` not modified.

## Verification notes

Each documented behaviour was read out of the source, and two of them contradicted the wording in the originating issue — worth a reviewer's attention, since the docs follow the source rather than the issue:

- `COMMANDS.md:97` — #1986 described `mf.force.kick` as "bypassing the normal role/permission check". It does not. `MfFactionKickCommand.kt` uses the permission only to pick which faction is resolved, then runs `faction.getRole(mfPlayer.id)` against that faction and refuses on `null`. The note documents the caveat; #1988 tracks the discrepancy.
- `COMMANDS.md:112` — same correction for `mf.force.rename` in `MfFactionSetNameCommand.kt`.
- `COMMANDS.md:363` — #1986 described `mf.force.unlock` as bypassing "another faction's" locked blocks. Locks are player-owned, not faction-owned, and both call sites are in the unlock path specifically. Corrected in a follow-up commit: only `PlayerInteractListener` sends the bypassed-owner message, `BlockBreakListener` unlocks silently, and the faction-role lock bypass applies to unlock mode only.
- `COMMANDS.md:126` — the force-flag note initially mentioned only `mf.flag.set`; corrected in the same follow-up, since `mf.flag.list` is likewise still required for the list path.
- `COMMANDS.md:500` — not from #1986: `MfFactionAddMemberCommand.kt:21` accepts `mf.force.addmember` **or** `mf.force.join`, but the existing entry listed only the former.
- All eight anchor links in the new table were checked against the actual `###` headings rather than assumed.

**Local build anchor:** UNVERIFIED-not-applicable — no Kotlin, resource or build file is touched, so `./gradlew test` has nothing to say about this diff. CI on the head SHA is the anchor, per the no-automated-coverage rule this does not overstate coverage of anything CI cannot reach.

## Deferred this cycle

Scoped to #1986. The rest of the open backlog is untouched, including the four other open PRs (#1915, #1918, #1944, #1948 — all currently conflicting with `main`). #1919 was groomed and self-reviewed in a previous cycle and is still awaiting human approval.


---

dmccoystephenson:
Two corrections to the line references in the self-review above, so they land on the right lines in the merged file: the `mf.force.kick` note is `COMMANDS.md:98` (not 97) and the `mf.force.unlock` note is `COMMANDS.md:368` (not 363). The other references (112, 126, 500) are correct.

On the `Closes #1986` caveat I raised: I went ahead and let it close. #1986's documentation half is fully delivered here, and its code half — the dead `mf.force.claim` / `mf.force.unclaim` declarations — is carried into **#1987** with the same evidence rather than dropped, which is what #1986's own "suggested scope" item 2 asked for. Nothing is lost by the closure.

Two follow-ups are now open and both need a maintainer decision rather than an agent guess:

- **#1987** — implement the admin claim/unclaim-for-another-faction feature, or delete the two dead permission declarations from `plugin.yml`.
- **#1988** — decide whether `mf.force.kick` / `mf.force.rename` should bypass the target faction's role check the way `mf.force.flag` does (a widening of an op-default permission), or whether the current targeting-only behaviour is intended and `plugin.yml`'s descriptions should be reworded instead.

The `COMMANDS.md` notes describe today's behaviour and point at each issue, so they need updating with whichever way those go.


---

Issue #1991, opened 2026-08-04
MfCuboidArea.distanceSquared overflows Int, letting gates be removed from far beyond gates.maxRemoveDistance


dmccoystephenson:
## Finding

`MfCuboidArea.distanceSquared` returns an `Int` and computes the squared distance with `Int` arithmetic:

```kotlin
fun distanceSquared(position: MfBlockPosition): Int {
    ...
    return ((closestX - x) * (closestX - x)) + ((closestY - y) * (closestY - y)) + ((closestZ - z) * (closestZ - z))
}
```

Squared distances above `Int.MAX_VALUE` (2,147,483,647) silently wrap around. The wrap begins at an axis offset of roughly 46,341 blocks, and the result can become negative. For example, a single-block area at `x = 60000` measured against a position at `x = 0` (same `y` and `z`) yields `60000 * 60000 = 3,600,000,000`, which wraps to `-694967296`.

## Impact

The only production caller is `MfGateRemoveCommand`:

```kotlin
val distanceSquared = gateDistanceSquared[closestGate]
val minRemovalDistanceSquared = plugin.config.getInt("gates.maxRemoveDistance") * plugin.config.getInt("gates.maxRemoveDistance")
if (closestGate == null || distanceSquared == null || distanceSquared > minRemovalDistanceSquared) {
    sender.sendMessage("$RED${plugin.language["CommandGateRemoveFailedToFindGate"]}")
    return@Runnable
}
```

When the wrapped value is negative, the `distanceSquared > minRemovalDistanceSquared` guard passes, so a gate can be deleted from a distance far greater than `gates.maxRemoveDistance` permits. Reaching this requires a player and one of their faction's gates to be separated by tens of thousands of blocks on a single axis, which is uncommon but reachable on large worlds and, on the far side of a Nether-style coordinate scale, on any world where a faction holds gates in widely separated regions.

`minByOrNull` over the same map is also affected: a wrapped-negative distance makes a very distant gate compare as the closest one.

## Current behaviour is characterized by a test

`MfCuboidAreaTest.testDistanceSquaredOverflowsForVeryDistantPositions` asserts the wrapped value so the behaviour is pinned rather than endorsed. That test must be updated when this is fixed.

## Suggested fix

[...]

Issue #1975, opened 2026-07-16
[Bug] server keeps randomly crashing and stop working


saito-samaaa:
<img width="1002" height="546" alt="Image" src="https://github.com/user-attachments/assets/48583908-73c1-40ff-a152-229c9772f867" />

<img width="997" height="729" alt="Image" src="https://github.com/user-attachments/assets/a5dc4c17-ef67-488d-a8b2-bef7945c66dd" />

## Description
i have this issue with medieval plugins tha keeps happening over and over it sometimes just randomly decides not to load and i would have to delete it and reinstall it for it to work it keeps happening often like almost  every other time i reopen my server, and sometimes even reinstalling doesn't work the first few times i have to do it 2 times+ for it to work i love this plugin so i hope there's a solution i'm on version 26.1.2 on papermc aternos

## Expected Behavior
the plugin works on server launch

## Actual Behavior
it crashes and i have to reinstall it

### Java Version:
- [ ] Java 25

### Plugin Version:
- [ ] v5.8.1

### Server Software & Version:
- [ ] Paper

Version: `_26.1.2_`

Issue #1971, opened 2026-04-26
[Feature] Lock picking


ninjakitty844:
## Is your feature request related to a problem?
*A clear and concise description of the problem you're facing.*
In castle sieges, even if you have managed to occupy and secure a significant portion of your enemy's castle, you still cannot take items from their locked chests or break through their locked doors in that area. And that is true EVEN IF chests and doors are listed in `wartimeInteractableBlocks` in the Medieval Factions configuration file.

You can steal claims from your enemy if you manage to lower their power below their number of claimed chunks, and then you would be able to freely interact with blocks in those claims. **But that is simply impossible to do if your enemy hides in their castle keep instead of fighting you down on the ground.** You won't be able to kill them to lower their power, because they are hiding behind locked doors you cannot unlock.

To be clear, I think it would be weird if your enemies could simply walk up and open doors or chests that are locked by your faction, which even _you_ (a regular faction member) might not be able to open on account of not being added as an accessor.

I think a much better solution would be to implement lock picking. This would also solve another issue I have with MF: there is little to no potential for thieves, since players can simply lock all of their doors/chests.

## Describe the solution you'd like
*A clear and concise description of what you want to happen.*

I have an idea for a lockpicking mechanic...

## Configuration options
(Note: This is just a mock up to the best of my ability. The option names and formatting could look different when implemented)

You should be able to define the starting strength of a lock depending on which type of block it is put on. For example, I should be able to give iron doors a starting lock strength of 20, and a wooden door a starting lock strength of 10. A lock strength of -1 would mean that the lock simply can never be lock picked.

`lockStrength:`
  - iron_door: 20
  - oak_door: 10

You should also be able to define the default starting lock strength, so that any block you didn't set a number for just starts at this default number.
`defaultLockStrength: 10`

In the configuration file, you should be able to define an item as a "lockpick":
`lockpickItem: "tripwire_hook"`

And you should also be able to toggle whether lockpicks are reusable:
`lockpickReusable: false`
If this is set to false, 1 lockpick should be removed from your hand after every lockpick attempt.

## Lockpicking process

Every time you break a locked block, the break should be cancelled like usual, but instead of just saying "That block is locked by [player_name].", the plugin should say "That block is locked by [player_name]. Lock strength: [remaining_lock_strength]."

[...]

Issue #1967, opened 2026-04-26
[Feature] Ability to use Minecraft's block tag feature to avoid redundant/excessive typing in config and allow easier customization


ninjakitty844:
## What are datapacks or block tags?

Minecraft has a feature called datapacks. These allow you to add/change things in Minecraft without actually altering its code or adding plugins. They can add recipes, run commands, and alter certain game behaviors. Datapacks have a feature called block tags, which are essentially a way to group block ids together so that you do not have to list every block id individually every time you want to perform an action on a group of blocks.

Here are some of Minecraft's default block tags (built-in):

**#minecraft:standing_signs**
`
{
  "values": [
    "minecraft:oak_sign",
    "minecraft:spruce_sign",
    "minecraft:birch_sign",
    "minecraft:acacia_sign",
    "minecraft:jungle_sign",
    "minecraft:dark_oak_sign",
    "minecraft:crimson_sign",
    "minecraft:warped_sign",
    "minecraft:mangrove_sign",
    "minecraft:bamboo_sign",
    "minecraft:cherry_sign"
  ]
}
`

**#minecraft:wall_signs**
`
{
  "values": [
    "minecraft:oak_wall_sign",
    "minecraft:spruce_wall_sign",
    "minecraft:birch_wall_sign",
    "minecraft:acacia_wall_sign",
    "minecraft:jungle_wall_sign",
    "minecraft:dark_oak_wall_sign",
    "minecraft:crimson_wall_sign",
    "minecraft:warped_wall_sign",
    "minecraft:mangrove_wall_sign",
    "minecraft:bamboo_wall_sign",
    "minecraft:cherry_wall_sign"
  ]
}
`

**#minecraft:signs**
`
{
  "values": [
    "#minecraft:standing_signs",
    "#minecraft:wall_signs"
  ]
}
`
**Note:** As seen here, block tags can contain other block tags! Convenient for organization :]

## Why should Medieval Factions respect block tags?

If you wanted to put all regular signs in your configuration file (for example, in the list of the blocks placeable during war), you would have to manually type out every single sign block id:
`
    "minecraft:oak_sign",
    "minecraft:spruce_sign",
    "minecraft:birch_sign",
    "minecraft:acacia_sign",
    "minecraft:jungle_sign",
    "minecraft:dark_oak_sign",
    "minecraft:crimson_sign",
    "minecraft:warped_sign",
    "minecraft:mangrove_sign",
    "minecraft:bamboo_sign",
[...]

PR #1992, opened 2026-08-04
test: characterize MfCuboidArea geometry behaviour

dmccoystephenson:
## Summary

A characterization test suite is added for `MfCuboidArea`, which had no test coverage despite backing three behaviours in the gate subsystem:

- **Gate creation validation** — `PlayerInteractListener` rejects a selection where `width > 1 && depth > 1` (flat-plane requirement) and where `height < gates.minHeight`.
- **Gate block enumeration** — `MfGate` opens and closes by iterating `area.blocks` and anchoring on `area.centerPosition`.
- **Gate removal** — `MfGateRemoveCommand` picks the closest gate with `area.distanceSquared(...)` and compares it against `gates.maxRemoveDistance` squared.

The behaviour pinned by these tests includes several conventions that are easy to break by accident:

- Corners may be supplied in any order; `minPosition` / `maxPosition` normalize per axis.
- `width` / `height` / `depth` are spans (`max - min`), not block counts — a single-block area has all three equal to `0`, and a 2×2 selection therefore passes the flat-plane check.
- `centerPosition` uses integer division, so it truncates towards zero (a negative midpoint rounds up).
- `blocks` is inclusive of both bounds and is ordered x, then y, then z.
- `contains` compares worlds; `distanceSquared` does not.
- Constructing an area from corners in two different worlds throws `IllegalStateException`.

One test, `testDistanceSquaredOverflowsForVeryDistantPositions`, pins current behaviour that is believed to be wrong: `distanceSquared` returns an `Int` and wraps to a negative value beyond roughly 46,341 blocks on an axis, which lets a gate be removed from far outside `gates.maxRemoveDistance`. Per the test-expansion convention, production code was not changed under a test cycle — the finding was filed separately as #1991 and the test is annotated to point at it.

## Test plan

- [x] New test file mirrors the structure and `// prepare` / `// execute` / `// verify` style of its siblings in `src/test/kotlin/com/dansplugins/factionsystem/area/`.
- [x] JUnit 5 via `kotlin.test`; no database, no network, no mocks required (`MfCuboidArea` is pure).
- [ ] `./gradlew clean test` — verified by CI on this PR's head commit. The local Gradle wrapper (7.6.4) cannot run in this environment against the installed JVM (`Unsupported class file major version 65`), so CI on the exact head SHA is the anchor, per the documented fallback.

## Data safety

No schema migrations; no DPC contract change; no config-default change; no production code change at all. The diff is a single new test file (+309 lines, 0 deletions).

## Backlog deferral note

This cycle selected proactive test expansion rather than an issue from the backlog. The remaining open issues were deferred for these reasons:

- **#1987, #1988** — both are explicitly filed as maintainer design calls (whether to implement or delete the dead `mf.force.claim` / `mf.force.unclaim` nodes; whether `mf.force.kick` / `mf.force.rename` should bypass the target faction's role check). Guessing either way would change permission semantics autonomously.
- **#1943, #1914, #1947, #1966, #1967, #1971, #1945, #1895, #1896, #1897, #1898, #1817, #1796, #1795, #1794, #1789, #1850, #1841, #1922, #1804, #1852** — feature requests or infrastructure work larger than a single polish-sized PR.
- **#1975, #1885, #1860, #1803, #1800, #1798, #1797** — bug reports lacking a reproduction path that can be confirmed from source alone.

The four open pull requests were also reviewed for grooming: #1944, #1918 and #1915 are all `CONFLICTING` against `main` at 3.5k / 1.2k / 2.1k lines respectively, and #1944 in particular has already been examined and deferred by four prior cycles for the same reason; #1948 was force-updated within the last day and is being actively iterated on, so it was left alone.

---

### DIFF
--- a/src/test/kotlin/com/dansplugins/factionsystem/area/MfCuboidAreaTest.kt
+++ b/src/test/kotlin/com/dansplugins/factionsystem/area/MfCuboidAreaTest.kt
@@ -1,0 +1,309 @@
+package com.dansplugins.factionsystem.area
+
+import java.util.*
+import kotlin.test.Test
+import kotlin.test.assertEquals
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertNotEquals
+import kotlin.test.assertTrue
+
+class MfCuboidAreaTest {
+
+    private val worldId: UUID = UUID.randomUUID()
+
+    private fun position(x: Int, y: Int, z: Int, world: UUID = worldId) = MfBlockPosition(world, x, y, z)
+
+    @Test
+    fun testInitializationRejectsMismatchedWorlds() {
+        // prepare
+        val otherWorldId = UUID.randomUUID()
+        val position1 = position(0, 64, 0)
+        val position2 = position(1, 64, 1, otherWorldId)
+
+        // execute
+        val exception = assertFailsWith<IllegalStateException> { MfCuboidArea(position1, position2) }
+
+        // verify
+        assertEquals("Position worlds do not match", exception.message)
+    }
+
+    @Test
+    fun testMinAndMaxPositionWhenCornersAreOrdered() {
+        // prepare
+        val position1 = position(0, 64, 0)
+        val position2 = position(4, 70, 8)
+
+        // execute
+        val area = MfCuboidArea(position1, position2)
+
+        // verify
+        assertEquals(position(0, 64, 0), area.minPosition)
+        assertEquals(position(4, 70, 8), area.maxPosition)
+    }
+
+    @Test
+    fun testMinAndMaxPositionNormalizeReversedCorners() {
+        // prepare - corners supplied in the opposite order
+        val position1 = position(4, 70, 8)
+        val position2 = position(0, 64, 0)
+
+        // execute
+        val area = MfCuboidArea(position1, position2)
+
+        // verify
+        assertEquals(position(0, 64, 0), area.minPosition)
+        assertEquals(position(4, 70, 8), area.maxPosition)
+    }
+
+    @Test
+    fun testMinAndMaxPositionNormalizePerAxis() {
+        // prepare - each axis is inverted independently
+        val position1 = position(4, 64, 8)
+        val position2 = position(0, 70, 0)
+
+        // execute
+        val area = MfCuboidArea(position1, position2)
+
+        // verify
+        assertEquals(position(0, 64, 0), area.minPosition)
+        assertEquals(position(4, 70, 8), area.maxPosition)
+    }
+
+    @Test
+    fun testDimensionsAreSpansNotBlockCounts() {
+        // prepare
+        val area = MfCuboidArea(position(0, 64, 0), position(4, 70, 8))
+
+        // execute / verify - width/height/depth are max - min, so they are one less than the block count
+        assertEquals(4, area.width)
+        assertEquals(6, area.height)
+        assertEquals(8, area.depth)
+    }
+
+    @Test
+    fun testDimensionsOfSingleBlockAreaAreZero() {
+        // prepare
+        val area = MfCuboidArea(position(10, 64, 10), position(10, 64, 10))
+
+        // execute / verify
+        assertEquals(0, area.width)
+        assertEquals(0, area.height)
+        assertEquals(0, area.depth)
+    }
+
+    @Test
+    fun testDimensionsAreOrderIndependent() {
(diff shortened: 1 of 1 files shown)


dmccoystephenson:
## Self-review rubric

Anchor: CI run [30875661482](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/30875661482) on head `be50b89e` — `build` pass, `docker-build` pass. The build log confirms `:lintKotlinTest`, `:test` and `:check` all executed and succeeded, so the anchor's scope does cover the one file this PR changes.

- **Scope: PASS** — `git diff --name-only origin/main...HEAD` returns exactly one path, `src/test/kotlin/com/dansplugins/factionsystem/area/MfCuboidAreaTest.kt`. No production file, no formatting churn, no unrelated rename.
- **Tests-new: N/A** — no new public method or function is introduced by this PR; the PR *is* the test addition. Coverage of the class under test is complete across its public surface: `minPosition`, `maxPosition`, `width`, `height`, `depth`, `centerPosition`, `blocks`, `contains`, `distanceSquared`, and the `init` world-mismatch guard each have at least one test.
- **Tests-fix: N/A** — no bug is fixed here. Per the test-expansion convention, the two defects found while writing these tests were filed (#1991, #1993) rather than patched under a test cycle, so there is nothing to stash-and-revert against.
- **Sibling structure: PASS** — the file matches `MfBlockPositionTest` / `MfChunkPositionTest` / `MfPositionTest`: same package, `kotlin.test` imports in the same order, `test<Behaviour>` naming, and the `// prepare` / `// execute` / `// verify` comment blocks. It deviates in not instantiating `TestUtils`, which is correct: `MfCuboidArea` is pure and needs no Bukkit mock.
- **Sibling renames: N/A** — no identifier renamed.
- **Docs: PASS, with a stated judgment call** — no documentation source of truth changes, because nothing user-facing changes. `CHANGELOG.md` was deliberately not touched: Keep a Changelog records changes visible to the people running the plugin, and a test-only addition is not one. Should the maintainers prefer test work to be logged, this is the point to say so.
- **Issue resolution: PASS** — no `Closes #N` is claimed. #1991 and #1993 are referenced from test comments and from the PR body specifically as *unfixed*, so neither is at risk of being auto-closed by this merge.
- **CI: PASS** — green on the exact head SHA, verified above.

### Repo-specific items

- **Migrations-additive: PASS** — `git diff origin/main...HEAD -- '*db/migration/*'` is empty. No migration file is added, edited, or deleted.
- **Lang-externalized: PASS** — no `.kt` or `.java` file under `src/main` is touched, so no user-facing string can have been added or hard-coded.
- **DPC-contract: PASS** — nothing under `dpc/` or `pacts/` appears in the diff.
- **Main-thread-safety: PASS** — the test file imports no Bukkit type at all (`java.util.UUID` and `kotlin.test` only), and `MfCuboidArea` itself touches no Bukkit API.
- **Plugin.yml: PASS** — no command or permission is added.
- **Config-defaults: PASS** — `src/main/resources/config.yml` is not in the diff.

### Findings worth a reviewer's judgment

These are not blocking, but an all-PASS rubric on a 300-line addition should not be taken at face value, and each of the following is a deliberate choice that could reasonably be made the other way.

- `src/test/kotlin/com/dansplugins/factionsystem/area/MfCuboidAreaTest.kt:275` — `testDistanceSquaredOverflowsForVeryDistantPositions` asserts the literal `-694967296`, which is a value the code should not be producing. When #1991 is fixed this test will fail, and the failure message ("expected -694967296") will read as though the fix is the regression. The comment above it links the issue, which is the intended mitigation, but `@Disabled("pinned pending #1991")` is a defensible alternative if the maintainers would rather the suite never assert a known-wrong result.
- `src/test/kotlin/com/dansplugins/factionsystem/area/MfCuboidAreaTest.kt:290` — `testEqualityDependsOnCornerOrderNotOnTheRegionCovered` pins `assertNotEquals` between two areas covering an identical region. That asymmetry is an artifact of the generated `data class` `equals`, not a designed property. Should anyone later give `MfCuboidArea` a normalizing `equals`, this test would flag that improvement as a regression. It is included because the current behaviour is load-bearing for anything that puts areas in a `Set` or uses one as a map key, but it is the weakest test in the file.
- The overflow threshold cited in the PR body and in #1991 (roughly 46,341 blocks on one axis) is derived arithmetically, not measured. The test itself uses a 60,000-block offset, which is comfortably past the boundary, so the assertion does not depend on the threshold being exact.

---

PR #1999, opened 2026-08-08
Resolve PlaceholderAPI placeholders in faction chat formats

### LINKED ISSUE #1994: Chat formats do not resolve PlaceholderAPI placeholders from other plugins
## Summary

Placeholders from other plugins are not resolved in Medieval Factions' configurable chat formats. A format string containing e.g. `%medievalroleplayengine_card_name%` is sent to recipients verbatim, with the placeholder text shown literally instead of its value.

This was reported on Discord by a server owner attempting to display Medieval Roleplay Engine character card names in faction chat.

## Reproduction

1. Install PlaceholderAPI and any plugin that registers an expansion (Medieval Roleplay Engine in the original report).
2. Add that plugin's placeholder to a chat format in `config.yml`:
   ```yaml
   chat:
     faction:
       format: '&7[faction] [${factionColor}${faction}&7] [${role}] &f%medievalroleplayengine_card_name%: ${message}'
   ```
3. Send a message in faction chat.

**Expected:** the placeholder is replaced with the sender's card name.
**Actual:** the literal text `%medievalroleplayengine_card_name%` appears in the message.

## Cause

`MfChatService.sendMessage` (`src/main/kotlin/com/dansplugins/factionsystem/chat/MfChatService.kt`) performs literal substitution of Medieval Factions' own `${...}` tokens and then translates colour codes. `PlaceholderAPI.setPlaceholders(...)` is not called there, nor anywhere else in the codebase.

Medieval Factions **registers** a PlaceholderAPI expansion in `MedievalFactions.kt` (`MedievalFactionsPlaceholderExpansion(this).register()`), so its own placeholders can be consumed by other plugins — which is why the reporter's TAB playerlist resolves MF placeholders correctly. The reverse direction is what is missing: Medieval Factions never resolves other plugins' placeholders inside its own config strings.

The reporter's own diagnosis ("maybe it is MF that doesn't support placeholders") was correct.

## Suggested approach

In `MfChatService.sendMessage`, when PlaceholderAPI is loaded, the assembled format should be passed through `PlaceholderAPI.setPlaceholders(bukkitPlayer, ...)` using the message sender as the player context. The call should be made **before** `translateAlternateColorCodes`, so that colour codes emitted by a placeholder are translated too.

PlaceholderAPI is already declared as a `softdepend` in `plugin.yml`, so the call must be guarded on the plugin being present, matching the existing guard used at registration time.

Two related notes:

- `sendMessage` is dispatched to the main thread by `AsyncPlayerChatListener`, so a synchronous `setPlaceholders` call is safe here.
- The default (non-channel) chat format in `AsyncPlayerChatListener` is hardcoded rather than read from config, so it has the same blind spot. Whether that path should also support placeholders is worth deciding as part of this issue.

---

dmccoystephenson:
## Summary

Placeholders belonging to other plugins are now resolved in Medieval Factions' configurable chat formats. Previously a format containing, for example, `%medievalroleplayengine_card_name%` was delivered to recipients verbatim, because `MfChatService.sendMessage` substituted only Medieval Factions' own `${...}` tokens and `PlaceholderAPI.setPlaceholders(...)` was never called anywhere in the codebase — Medieval Factions registered an expansion but never consumed one.

Closes #1994

## What was changed

- A small seam, `MfPlaceholderResolver`, was added, along with its PlaceholderAPI-backed implementation `MfPlaceholderApiResolver`. The implementation is guarded on `plugin.server.pluginManager.getPlugin("PlaceholderAPI") != null`, matching the guard already used where the expansion is registered in `MedievalFactions.kt`. When PlaceholderAPI is not installed, the text is returned unchanged and the PlaceholderAPI class is never touched.
- The message assembly in `MfChatService` was extracted into `formatMessage(...)`, and the raw format template is now passed through the resolver using the **message sender** as the player context.
- `MfChatService` takes the resolver as a third constructor parameter defaulting to `MfPlaceholderApiResolver(plugin)`, so the existing call site in `MedievalFactions.kt` was left untouched and the seam stays injectable for tests.
- The fallback used when no format is configured is now selected from the `channel` parameter rather than from `mfPlayer.chatChannel`. The two are the same value at the only call site (`AsyncPlayerChatListener` passes `mfPlayer.chatChannel`), so no behaviour changes on a live server; what it removes is the unreachable `null ->` branch that previously produced an empty message. This was raised in review and is included here for completeness.
- Documentation was updated: a `### PlaceholderAPI placeholders in chat formats` note was added under Chat Settings in `CONFIG.md`, and an entry was added to `CHANGELOG.md` under `## [Unreleased]` → `### Added`.

## Ordering, and the injection hazard it closes

The order of operations in `formatMessage` is a correctness constraint, not a preference, and it is documented in KDoc and in an inline comment at the call site so that a future refactor is not able to undo it silently:

1. **Placeholders are resolved against the raw format template**, before anything else.
2. Medieval Factions' own `${...}` tokens are substituted.
3. Colour codes are translated.

An earlier revision of this branch performed step 1 *after* step 2. That was reported during review and has been fixed. Under that ordering, player-controlled text was being handed to `PlaceholderAPI.setPlaceholders`: a player typing `%some_placeholder%` into faction chat would have had it expanded against their own context, as would a `%`-bearing display name set by a nickname plugin, since `${message}` and `${displayName}` were already substituted by that point. Only registered expansions are expanded by PlaceholderAPI, so the blast radius depended on what a given server had installed — which is precisely the hazard, given expansions such as JavascriptExpansion or any expansion exposing sensitive or expensive data. With resolution moved ahead of substitution, only the server owner's own config string is ever passed to the placeholder provider.

Colour code translation was kept last, so colour codes emitted by a placeholder are still translated.

One consequence of the new order is worth naming: a value returned by a placeholder is itself scanned for `${...}` tokens afterwards, so a placeholder resolving to, say, the literal text `${faction}` would have that token substituted. This is considered acceptable — the substituted values are the faction name, role, player name and the sender's own message, none of which are privileged, and the alternative (a single-pass substitution) would add complexity for no security benefit. It is called out here so that the trade-off is a decision rather than an oversight.

## Dependency declaration

No build change was required. `me.clip:placeholderapi:2.11.6` is already declared in `build.gradle`, in the same way as the plugin's other soft dependencies (`spigot-api`, `dynmap-api`), and it is deliberately **not** listed in the `shadowJar` `dependencies { include(...) }` allowlist, so it is neither bundled nor shaded. `compileOnly` would arguably be more precise, but `implementation` is the convention used throughout this file and switching only this one entry would be inconsistent; the shipped jar is unaffected either way.

## The default (non-channel) chat format

The hardcoded default format in `AsyncPlayerChatListener` was deliberately left alone. Two reasons are offered for review:

1. That format is not config-driven — it is built in code from `ChatColor` values and the faction prefix, so it contains no placeholder text for PlaceholderAPI to resolve. Passing it through the resolver would be a no-op today.
2. It is assigned to `event.format`, which Bukkit later feeds to `String.format`. Injecting placeholder output into it would risk a `%` in a resolved value being interpreted as a format specifier.

Making that format configurable is a separate feature, and it would inherit placeholder support for free once the format string is read from config and run through the same path.

## Testing

New coverage:

- `MfPlaceholderApiResolverTest` (3 tests) — PlaceholderAPI absent returns the text unchanged and does not touch the static API (asserted with `MockedStatic.verifyNoInteractions()`); PlaceholderAPI present returns the resolved text. `Mockito.mockStatic` is used, following the pattern already established in `CreatureSpawnListenerTest`.
- `MfChatServiceTest` (6 tests) — only the configured format string is handed to the resolver; the resolver is called with the message sender as the player context; **a `%...%` sequence typed into the chat message body is not expanded**; **a `%...%` sequence in a display name is not expanded**; colour codes emitted by a placeholder are translated in the final message; and a resolver that changes nothing still produces the previously expected formatting.

Both ordering directions were re-confirmed empirically on the rebased head: with the resolver call deleted, three of the six `MfChatServiceTest` cases fail; with the resolver call moved back below the `${...}` substitutions, the two injection cases and the template-only case fail. The suite passes again once the call is restored to its committed position.

What is *not* covered by unit tests is the real end-to-end path with a live PlaceholderAPI instance. That was covered separately by the in-game verification recorded in a comment on this pull request, run on Spigot 1.21.1 with PlaceholderAPI 2.11.6 against both `main` and this branch.

## Update — rebased on `main`

The branch was 24 commits behind `main` (the JSON storage backend, the dev-release workflow and the gate-distance fix have landed since). It has been rebased onto `cc688dab` and force-pushed; the rebase was clean, with only `CHANGELOG.md` and `CONFIG.md` auto-merging.

Data safety: no Flyway migration is added or edited, `src/main/resources/config.yml`, `plugin.yml` and the `lang/` files are untouched, nothing under `dpc/` is changed, and no config default is altered — only `CONFIG.md`'s prose. The change is confined to chat message formatting.

Verification on the rebased head: `./gradlew clean test` and `./gradlew lintKotlin` both pass locally, and the CI `Build` workflow (`build`, `changes` and `docker-build` jobs) passes on the pushed head. Earlier CI runs on this branch had been left in `action_required` because the previous head commit was pushed by a bot account, which is why no green check was showing before the rebase.

*Implemented by Claude on behalf of Daniel Stephenson.*

---


dmccoystephenson:
## Verified in-game

This was tested on a live Spigot 1.21.1 server with PlaceholderAPI 2.11.6 installed and a real client joined, rather than by unit tests alone. The same scenario was run twice: once against `origin/main` (`abeab932`) and once against this branch's head (`79b4f9e7`).

`chat.faction.format` was set to:

```yaml
format: '&7[faction] [papi:%MedievalFactions_faction_name%] &f${displayName}: ${message}'
```

and the player (a member of faction `Placeholderia`) then sent a message whose **body** contained that same placeholder token.

| Build | Line delivered to the channel |
| --- | --- |
| `origin/main` `abeab932` | `[faction] [papi:%MedievalFactions_faction_name%] Alice: hello %MedievalFactions_faction_name% world` |
| this PR `79b4f9e7` | `[faction] [papi:Placeholderia] Alice: hello %MedievalFactions_faction_name% world` |

Both halves of the change are demonstrated by that single line:

- **The reported bug is fixed.** The placeholder written into the format by the server owner is resolved; on `main` it is rendered literally, which is exactly what was reported.
- **The injection guard holds on a live server.** The identical token, typed by the player into chat, survives verbatim in the message body. Had the resolution been ordered after `${message}` substitution, it would have expanded there too.

That the placeholder itself was resolvable was confirmed independently beforehand via `papi parse Alice %MedievalFactions_faction_name%` → `Placeholderia`.

### Scope of this verification

The expansion exercised was Medieval Factions' own, not MedievalRoleplayEngine's, which was not installed. It is registered with PlaceholderAPI through the same public registry and resolved through the same `PlaceholderAPI.setPlaceholders` call, so the code path is identical — but the originally reported plugin was not itself present.

The behaviour of `AsyncPlayerChatListener`'s default format was not exercised, since it is deliberately left unchanged by this PR.

---

dmccoystephenson:
## Review of the rebased head (`c8d3498`)

This branch was rebased onto `main` (`cc688dab`) during this session and re-reviewed in full afterwards, since the previous review predates 24 commits of `main` — including the JSON storage backend, which introduces a second storage path that this change had never been read against.

### Rubric

- **CI:** PASS — `build`, `changes` and `docker-build` all pass on `c8d3498`. Earlier runs on this branch sat in `action_required` because the prior head was pushed by a bot account; the rebase push re-triggered them under a human account and they completed green.
- **Local anchor:** PASS — `./gradlew clean test` and `./gradlew lintKotlin` both succeed on the rebased head.
- **Scope:** PASS — seven files are touched (`CHANGELOG.md`, `CONFIG.md`, `MfChatService.kt`, the two new `placeholder/` classes, and two test classes). No unrelated formatting, rename or comment churn appears in `git diff origin/main`.
- **Tests-new:** PASS — both new public surfaces are exercised: `MfChatService.formatMessage` by six cases and `MfPlaceholderApiResolver.resolve` by three, including the PlaceholderAPI-absent path asserted with `MockedStatic.verifyNoInteractions()`.
- **Tests-fix (empirical):** PASS — verified by mutation rather than by reasoning. With `MfChatService.kt:96` deleted, `formatMessage_ShouldPassOnlyTheConfiguredFormatToThePlaceholderResolver`, `formatMessage_ShouldResolvePlaceholdersAgainstTheMessageSender` and `formatMessage_ShouldTranslateColourCodesEmittedByAPlaceholder` fail. With the same call moved below the `${...}` substitutions instead, `formatMessage_PlaceholderTypedIntoTheMessage_ShouldNotBeResolved`, `formatMessage_PlaceholderInADisplayName_ShouldNotBeResolved` and the template-only case fail. The suite passes again with the call restored, and the tree was left clean.
- **Migrations-additive:** PASS — `git diff origin/main -- '*db/migration/*'` is empty. No migration is added, and no released migration is edited.
- **Lang-externalized:** PASS — no user-facing string literal is added to any `.kt` file; the additions are KDoc and one inline comment. No `lang/` file is touched, so no existing key or translation is removed.
- **DPC-contract:** PASS — nothing under `dpc/` or `pacts/` appears in the diff, so the `POST /api/v1/factions` snapshot is unaffected.
- **Main-thread-safety:** PASS — the only caller is `AsyncPlayerChatListener.kt:36`, which dispatches `sendMessage` through `scheduler.runTask`. `PlaceholderAPI.setPlaceholders` is therefore invoked on the main thread, which is what third-party expansions expect; no new Bukkit access is introduced on an async thread.
- **Config-defaults / plugin.yml:** PASS — `src/main/resources/config.yml` and `plugin.yml` are untouched. Only `CONFIG.md` prose is added, describing keys that already exist.
- **Docs:** PASS — `CHANGELOG.md` carries an `## [Unreleased]` → `### Added` entry, and `CONFIG.md:507` documents the behaviour under Chat Settings next to the three format keys it applies to. `COMMANDS.md`, `FACTION_FLAGS.md` and `DATABASE_QUERYING.md` are not implicated; `DEPENDENCY_USAGE.md` documents consuming Medieval Factions as a dependency rather than its own dependencies, so it needs no entry.
- **Sibling structure:** PASS — the new classes join the existing `placeholder/` package beside `MedievalFactionsPlaceholderExpansion.kt`, and the new tests mirror the JUnit 5 / Mockito `mockStatic` shape already used by `MedievalFactionsPlaceholderExpansionTest` and `CreatureSpawnListenerTest`.
- **Issue resolution:** PASS — #1994 reports that a placeholder written into `chat.faction.format` is delivered verbatim; that path is the one changed, and the in-game record already posted on this pull request shows the token resolving on this branch and not on `main`.

### Notes, none of them blocking

- `src/main/kotlin/com/dansplugins/factionsystem/chat/MfChatService.kt:96` — the value returned by an expansion is scanned for `${...}` tokens by the six `replace` calls that follow it. A placeholder resolving to the literal text `${faction}` would therefore be substituted. The pull request body names this and accepts it; it is repeated here only so the trade-off is visible in review as well. The substituted values are not privileged.
- `src/main/kotlin/com/dansplugins/factionsystem/chat/MfChatService.kt:88` — the fallback branch is now chosen from the `channel` parameter rather than `mfPlayer.chatChannel`. Since the sole caller passes `mfPlayer.chatChannel`, this is behaviour-neutral today and removes an unreachable `null` branch; it is worth being aware of should a second caller ever pass a channel other than the sender's own, as the fallback would then follow the destination channel rather than the sender's selection. That is the more defensible of the two behaviours.
- `src/test/kotlin/com/dansplugins/factionsystem/chat/MfChatServiceTest.kt` — coverage stops at `formatMessage`. The recipient fan-out, the asynchronous repository insert and the console echo in `sendMessage` remain untested. This is pre-existing and out of scope here, but it means the extracted method, not the sent message, is what these tests protect.
- No unit test exercises a real PlaceholderAPI instance; the seam is mocked throughout. The in-game verification comment above is what covers that gap, and it exercised Medieval Factions' own expansion rather than the third-party one from the original report.

### Data safety

No schema migration, no DPC contract change, no `config.yml` default change, no `lang/` key removal, no build-file change. Nothing in this diff can reach persisted faction, claim, player or power data — the change is confined to formatting a chat line before it is sent.

This review was performed during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

Issue #1747, opened 2023-07-14
Can't eat food while looking at a different faction's blocks


ninjakitty844:
## Describe the bug
If you are looking at a different faction's claimed blocks, you cannot eat.

## To Reproduce
Look at a different faction's claimed blocks. Try to eat.

## Expected behavior
I would expect to just be able to eat.

## Screenshots or Logs
(If applicable, add screenshots to help explain your problem.)

## Java version
 - [ ] Java 8
 - [ ] Java 11
 - [ ] Java 16
 - [x ] Java 17 

**Plugin version**
 - [ ] v4.1
 - [ ] v4.2
 - [ ] v4.3
 - [ ] v4.4
 - [ ] v4.5.x
 - [x ] v5.20
---
**Spigot/Paper version** _(Please Specify which server software you are using)_
 - [ ] 1.16.x
 - [ ] 1.17.x
 - [ ] 1.18.x
 - [x ] 1.19.3

dmccoystephenson:
I remember this being an issue in MF 4. Previous attempts to fix it introduced a new bug that allowed players to bypass protections when holding certain items. We need a solution that prevents this exploit. One possibility is canceling the interaction event while simulating the eating action.

PR #2000, opened 2026-08-08
Protect only the block half of a denied interaction

### LINKED ISSUE #1995: Non-edible hand-used items (bows, snowballs, ender pearls) are still blocked when looking at a protected block
## Summary

The fix that allowed food to be eaten while looking at a protected block whitelists edible items only. Every other item that is used from the hand — bows, snowballs, ender pearls, fishing rods, tridents, splash potions, eggs — is still blocked when the player happens to be looking at a block inside a claim they cannot interact with.

This was reported on Discord as a follow-up to the eating fix.

## Current behaviour

In `PlayerInteractListener` (`src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt`):

```kotlin
// Allow eating food in faction territory without triggering protection
val item = event.item
if (item != null && item.type.isEdible && !clickedBlock.type.isInteractable) {
    return
}
```

Anything failing `isEdible` falls through to `event.isCancelled = true` further down, which cancels the interaction entirely.

## Why the whitelist is the wrong shape

`PlayerInteractEvent` carries two independent results — one for the interacted block and one for the item in hand. Setting `isCancelled = true` denies **both**, which is why using an item is suppressed as a side effect of protecting the block. Extending the edible check into a growing list of allowed materials would treat the symptom.

Denying only the block half expresses the actual intent:

```kotlin
event.setUseInteractedBlock(Event.Result.DENY)
```

This covers every self-use item at once with no material list to maintain, and it allows the `isEdible` special case to be **removed** rather than extended.

Block placement is not opened up by this change: `BlockPlaceListener` performs its own independent claim check (`claimService.isInteractionAllowed(...)`, plus the wartime placeable-block allowances), so a placement attempt is still rejected there.

## Interaction with other plugins

A related concern was raised in the same report: actions contributed by other plugins (the example given was right-clicking an apple on a cauldron to add it to a brewing mix) should remain blocked.

That case is already handled and is not regressed by this change — `CAULDRON` is interactable, so the edible bypass never applied to it, and denying the block result keeps it denied. Switching to `setUseInteractedBlock(DENY)` is in fact an improvement here, because well-behaved plugins that check `event.useInteractedBlock()` will see the denial rather than seeing an uncancelled event.

Plugins that listen to `PlayerInteractEvent` and ignore the result fields cannot be stopped from within Medieval Factions. That limit is inherent to the Bukkit API and should not be promised away in any response to the reporter.

## Acceptance

- Bows, snowballs, ender pearls, fishing rods and similar hand-used items function while a protected block is being looked at.
- Chests, doors, buttons, cauldrons and other interactable blocks remain protected.
- Blocks cannot be placed in a claim the player may not interact with.
- The existing wartime ladder / placeable / breakable / interactable allowances continue to behave as before.
- `PlayerInteractListenerTest` is extended to cover the non-edible item cases.

---

dmccoystephenson:
## Summary

Non-edible hand-used items were still being suppressed when a player was looking at a protected block. A bow, crossbow, snowball, egg, ender pearl, fishing rod, trident, splash potion, shield, spyglass or goat horn could not be used at all in another faction's claim, in wilderness with `wilderness.interaction.prevent` enabled, or at a locked block.

The cause is that `event.isCancelled = true` denies **both** of `PlayerInteractEvent`'s results — the interacted block and the item in hand. Denying the item in hand also makes CraftBukkit swallow the follow-up use-item packet that the client sends when a block-targeted use does nothing, which is the packet that draws a bow, throws a snowball, casts a rod, or starts eating. Protection therefore reached far past the block it was protecting. The consumable-material whitelist added for #1747 (`isConsumingItem` / `isConsumableItem` / `DRINKABLE_MATERIALS`) worked around this for food and three drinkables only.

Interaction protection is now expressed as denying the interacted-block result alone whenever the held item's use cannot act on the block. The three whitelist members named in the issue have been removed and replaced by a single `denyInteraction` helper used by all three protection paths.

## Divergence from the proposed fix

The issue proposed replacing `isCancelled = true` with `setUseInteractedBlock(Event.Result.DENY)` unconditionally, and deleting the material list outright. That was followed in shape but **not** unconditionally, for two reasons that were verified against the code:

1. **Releasing the item half for every item opens real griefing holes.** An item's block-targeted use (`ItemStack.useOn`) is gated on the *item* result, not the block result. Releasing it lets a non-member ignite a claim with flint and steel, till it with a hoe, strip its logs with an axe, path it with a shovel, bone-meal it, or place an item frame, painting, armour stand, boat, minecart or end crystal in it. `BlockPlaceListener` and `PlayerBucketListener` cover placement and buckets, but no listener in this plugin handles `BlockIgniteEvent`, `EntityPlaceEvent`, `HangingPlaceEvent` or `BlockFertilizeEvent`, so cancellation of the interact event is currently the only thing stopping those. Any list drawn up here has to be **fail-closed** — an item missing from it must be merely restricted, never a protection hole — so a blacklist of block-acting items was rejected in favour of an allowance list of items that provably have no block-targeted use. The list is now much wider than "consumable" and no longer needs extending per item class for the common cases, but it is deliberately retained. That underlying gap is tracked separately as #2002.
2. **A left-click and a physical interaction are gated on `isCancelled` alone.** Block-break attempts and pressure-plate/farmland/tripwire interactions are checked by the server against the cancelled flag, not against the block result, so denying only the block result there would silently stop protecting them. Those paths are still cancelled in full.

One further improvement falls out of the new shape: because the block is now protected by the block result rather than by an early return, the exemption no longer has to be withheld from interactive blocks. Eating, or drawing a bow, while facing a chest is allowed, and the chest still stays shut.

## The `lockedBlock` path

It was changed too, deliberately. A lock is a separate protection mechanism from territory, but the two-result distinction applies to it identically: the locked block must stay shut, while an item that cannot act on it has no reason to be suppressed. Being unable to eat or shoot while standing at someone else's locked chest is the same bug in a different place. The change is safe there specifically because the item half is only released for items with no block-targeted use — the lock path `return`s and thereby skips territory protection, so releasing a block-acting item there would have bypassed two protections rather than one.

## What a server owner will observe differently after upgrading

- In another faction's claim, in wilderness with `wilderness.interaction.prevent` enabled, and at another player's locked block, players can now **draw a bow or crossbow, throw a snowball, egg, ender pearl, trident or splash potion, cast a fishing rod, raise a shield, and use a spyglass or goat horn** while looking at a protected block. Previously all of these did nothing.
- **Splash and lingering potions are now thrown** rather than suppressed. They were previously excluded from the eat/drink exemption on the grounds of being thrown rather than drunk; throwing one still cannot alter a block, and the harm it can do is already protected against by `PotionSplashListener`, `LingeringPotionSplashListener` and `AreaEffectCloudApplyListener`.
- **Right-clicking an interactive block while holding one of those items no longer produces the protection message**, and no longer opens the block either. The block half is refused silently and the item is used. The message is now sent only when the interaction is refused in full, which is what keeps #1747's chat noise from returning.
- Nothing else is loosened. Chests, levers, doors, buttons and cauldrons stay protected whatever is held; block placement, bucket use, flint and steel, hoes, axes, shovels, bone meal, eyes of ender, firework rockets and wind charges are still refused outright; left-clicks and physical interactions are still cancelled in full; and the wartime ladder, placeable, breakable and interactable allowances are untouched.

## Tests

`PlayerInteractListenerTest` was extended with the items named in the issue (bow, snowball, ender pearl, egg, fishing rod, trident, splash potion) and with a test per invariant:

- a left-click with food, and with a bow, is still cancelled in full and still notifies;
- a physical interaction with a bow in hand is still cancelled in full, and still without a message;
- an interactive block has its block half denied with food, and with a bow, in hand;
- flint and steel, and a placeable block, are still cancelled in full — the two cases that justify keeping an allowance list;
- a bow at a locked block denies the block half, allows the item, and schedules no owner lookup;
- a bow in wilderness with `wilderness.interaction.prevent` enabled is allowed.

`BlockPlaceListenerTest` gained the missing case confirming that placement in a claim the player may not interact with is refused independently by `BlockPlaceListener`.

Two existing tests changed meaning and were rewritten rather than deleted: food against an interactive block (now block-denied rather than cancelled) and a splash potion (now allowed). New verifiers `verifyBlockUseDenied` and `verifyItemUseNotDenied` assert the two halves separately, the latter covering both routes by which the item half can be denied.

## Grooming performed in a later cycle

The branch was carried forward and brought to a mergeable state:

- It was rebased onto `main`, which it was 31 commits behind (the JSON storage backend, the dev-release workflow, the gate-distance fix and the merge of #1999). The rebase produced no conflicts outside a trivial `CHANGELOG.md` adjacency.
- The `build` check had been failing on `lintKotlinMain`, not on a test: ktlint's `wrapping` rule rejected the call wrapping of the new material list at `PlayerInteractListener.kt:60` and `:76`. The list is now wrapped as ktlint requires. No logic was altered by that commit.
- `CONFIG.md` was found to still describe the superseded consumable-only exemption under `wilderness.interaction.prevent`, and to omit the new message-suppression condition under `wilderness.interaction.alert`. Both entries were rewritten to match the implementation.

The regression gate was checked empirically rather than by inspection: with `PlayerInteractListener.kt` reverted to its `origin/main` content and the tests left in place, 12 of the new tests fail; with the fix restored, all 66 pass.

**Test plan**

- [x] `./gradlew lintKotlin test` passes locally on the rebased head
- [x] `build`, `changes` and `docker-build` pass in CI on the head SHA
- [x] The 12 new tests are confirmed to fail with the fix reverted and pass with it restored
- [x] Verified in-game on Spigot 1.21.1 against `origin/main` as a control (see the earlier comment on this pull request)

**Data safety:** no schema migrations; no DPC contract change; no `config.yml`, `plugin.yml` or `lang/` change. The modified paths are one listener, two test files, `CHANGELOG.md` and `CONFIG.md`.

Closes #1995


dmccoystephenson:
## Verified in-game

This was tested on a live Spigot 1.21.1 server with a real client joined as a non-member standing inside another faction's claim. Each case reproduces a vanilla right-click faithfully — `use_item_on_block` followed by `use_item` in the same tick, with the server-side raytrace landing on the same block — because CraftBukkit only suppresses the follow-up packet when its own raytrace agrees, and a bot that is not genuinely aiming at the target will not reproduce the bug at all. That the sequence was faithful was confirmed first as a control: on `origin/main` the vanilla sequence produced no snowball, while an open-air control did.

The identical script was then run against `origin/main` (`abeab932`) and this branch's head (`47e66e90`).

**Baseline:** all 13 cases denied, every one with a protection message. The reported bug reproduces.

**This PR**, non-member inside the claim:

| Item held | Item use runs | Block stays protected | Protection message |
| --- | --- | --- | --- |
| snowball, egg, ender pearl, experience bottle | yes — projectile spawned, 1 consumed | yes — chest never opened, 0/8 attempts | none |
| **splash potion, lingering potion** | **yes** | yes | **yes — see below** |
| flint and steel | no — no fire block appeared | yes | yes |
| bone meal, item frame, water bucket, oak boat | no — nothing consumed, no entity spawned | yes | yes |
| bare hand at a chest | — | yes — chest stayed shut | yes |

So the reported behaviour is fixed, the two-result split works as designed, and the fail-closed allowance list holds: nothing that could place, ignite or alter a block got through, while the chest remained shut in every single case including those where the held item was released.

The design decision that no message is sent when only the block half is denied is confirmed in practice: right-clicking a chest with a snowball throws the snowball and says nothing.

### One finding worth a decision

**Splash and lingering potions are thrown successfully and the thrower is simultaneously told they cannot interact.** This reproduced on 8 of 8 attempts, against both an interactive block (chest) and a plain one, while the four other throwables in the allowance list never produced it.

```
item                target    used  spawned                    warned
snowball            chest     1     snowball                   false
egg                 chest     1     egg                        false
ender_pearl         chest     1     ender_pearl                false
experience_bottle   chest     1     experience_bottle          false
splash_potion       chest     1     potion                     true
lingering_potion    chest     1     potion,area_effect_cloud   true
```

`CannotInteractWithBlockInFactionTerritory` has exactly one sender in the codebase (`PlayerInteractListener.kt:283`, gated on `deniedInFull`), so a second `RIGHT_CLICK_BLOCK` event must be firing and being denied in full. The distinguishing property of the two affected items is that they carry data components (`potion_contents`); the likely mechanism is CraftBukkit's `handleUseItem` failing its cached-interact comparison for such a stack and re-firing the event rather than reusing the cached result. **This was not confirmed**, and it may well be pre-existing behaviour rather than something this PR introduces — but the user-visible result is that an action which succeeds is also denounced, which is the mirror image of the silent-denial trade-off already documented here.

### Not covered

Bow draw, drinking and eating could not be decided by this rig: the bot produced nothing for those items **even in unclaimed wilderness with no protection configured**, so a null result says nothing about this change. The `#1747` eating regression therefore remains unverified in-game, as does the bow case from the original report. Only instantaneous item uses are decidable here.

---

dmccoystephenson:
## Status note from this cycle

This pull request was not the one groomed this cycle; #1999 was taken first as the oldest of the three, rebased onto `main` and merged. One pull request is groomed per cycle, so this one is next in line rather than deferred on its merits.

Two things are worth recording so that the next pass does not have to rediscover them:

- The branch is now behind `main` by the JSON storage backend, the dev-release workflow, the gate-distance fix and the merge of #1999. A rebase is needed before it can be assessed.
- No check is currently reported on this pull request. The cause is not a failure: the head commit was pushed by the Copilot agent while addressing review comments, so the required `Build` workflow is held in `action_required` and the pull request reads as `BLOCKED`. A rebase and force-push from a human account re-triggers it, which is what unblocked #1999.

The finding already recorded above — that splash and lingering potions are thrown successfully while their thrower is simultaneously told the interaction is denied — remains open and is a maintainer decision rather than something this loop will resolve autonomously.

The wider open-issue backlog was likewise not picked up this cycle: the cycle was scoped to bringing an existing pull request to a mergeable state, which the skill treats as a complete unit of work. No judgement about the remaining issues is implied by their being untouched.

This comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

dmccoystephenson:
## Review of the rebased head (`344e5074`)

The diff was read in full and scored against this loop's rubric. Each verdict is grounded in the diff or in a command output rather than in judgement.

### Rubric

- **Scope: PASS** — five files are touched (`PlayerInteractListener.kt`, `PlayerInteractListenerTest.kt`, `BlockPlaceListenerTest.kt`, `CHANGELOG.md`, `CONFIG.md`) and each is required by #1995. Net non-test change is roughly 50 lines, well inside the scope ceiling. No unrelated formatting or comment churn was found.
- **Tests-new: PASS** — the one new production member, `denyInteraction`, is exercised through both of its outcomes: block-half-only denial (bow, snowball, egg, ender pearl, trident, fishing rod, splash potion, food) and full cancellation (flint and steel, a placeable block, left-click, physical).
- **Tests-fix: PASS, verified empirically rather than reasoned.** `PlayerInteractListener.kt` was reverted to its `origin/main` content with the tests left in place; 12 of the new tests failed. With the fix restored, all 66 passed. The failures were observed, not inferred.
- **Migrations-additive: PASS** — `git diff origin/main...HEAD -- '*db/migration/*'` is empty. No schema change of any kind.
- **Lang-externalized: PASS** — no string literal is added to the changed Kotlin. Every message on the changed paths continues to go through `plugin.language[...]`; the only change to messaging is *whether* an existing key is sent.
- **DPC-contract: PASS** — nothing under `dpc/` or `pacts/` is touched.
- **Main-thread-safety: PASS** — no async dispatch is added. The one async block on a changed path (the locked-block owner lookup) is pre-existing and is only newly *gated*, never newly scheduled.
- **Plugin.yml / Config-defaults: PASS** — neither `plugin.yml` nor `config.yml` is modified, so no default silently changes on an existing install. This is what keeps the pull request off the do-not-auto-merge list.
- **Docs: PASS after a fix made this cycle.** `CONFIG.md:184` still described the superseded consumable-only exemption, and `CONFIG.md:189` omitted the new condition under which no wilderness alert is sent. Both were rewritten in `344e5074`. `CHANGELOG.md` carries the change under `### Fixed`. `COMMANDS.md`, `FACTION_FLAGS.md`, `DATABASE_QUERYING.md` and `DEPENDENCY_USAGE.md` are not implicated.
- **Issue resolution: PASS** — all four acceptance criteria in #1995 are met, and the one divergence from the issue's proposed fix (an allowance list is retained rather than removed) is argued in the description and tracked as #2002.
- **CI: PASS** — `build`, `changes` and `docker-build` are green on `344e5074`. The previous failure was `lintKotlinMain` rejecting the call wrapping of the new material list, not a test; it has been corrected.

### Findings folded in from the diff

Anchored inline comments are unavailable in this session, so the three observations below are recorded here as `path:line — finding`.

- `src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:322` — **`item.type.isEdible` is not quite the world-neutral predicate the name promises.** Sweet berries and glow berries are reported as edible by Bukkit and *do* have a block-targeted use: they can be planted. The item half is therefore released for them. This is not a regression — the previous `isConsumingItem` early return released the entire event for the same materials, so the exposure is strictly narrower now than before — and `BlockPlaceListener` refuses the resulting placement independently, which was confirmed by reading its `isInteractionAllowed` check and is now covered by the new `BlockPlaceListenerTest` case. Raised because the `isEdible` half of the allowance rests on that second barrier while the explicit half does not, which is precisely the fragility #2002 describes.
- `src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:73` — **`Material.SHIELD` has no effect in the allowance list in the usual case.** A shield is ordinarily held in the off hand, and `onPlayerInteract` returns early on `event.hand != HAND`, so this listener never sees that interaction at all. The entry is harmless and correct for a main-hand shield, but it should not be read as the thing that fixes off-hand shield raising, which was never broken.
- `src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractListener.kt:214` — **the locked-block path now stays silent for an allowed item.** A non-accessor holding a bow at someone's locked chest is told nothing, where previously a lock message was sent. This is consistent with the territory path and with the intent of #1747, and the chest still does not open, but it is a user-visible change on the lock path that the changelog entry describes only in terms of claims.

### Not resolved by this pull request

The finding recorded earlier on this pull request — that splash and lingering potions are thrown successfully *and* their thrower is simultaneously told the interaction is denied — was re-examined against the diff and is not addressed here. `CannotInteractWithBlockInFactionTerritory` has a single sender, gated on `deniedInFull`, so a second `RIGHT_CLICK_BLOCK` event must be firing for those two stacks and being denied in full. The effect is a spurious message; the block is still protected in that case, so it is cosmetic rather than a protection hole. It would be better tracked as its own issue than held against this change.

### Verdict

Green on every rubric item. The protection surface is narrowed nowhere that was verified: every path that could place, break, ignite or alter a block is still cancelled in full, the two paths gated on `isCancelled` alone are untouched, and the block half is denied explicitly on every path that previously cancelled. Recommended for merge.

This review was performed and posted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

dmccoystephenson:
## Cycle record

This pull request was the one groomed this cycle, as the oldest non-draft of the three carried forward. Grooming consisted of a rebase onto `main`, the ktlint fix that had been failing the required `build` check, a `CONFIG.md` correction found by the documentation pass, an empirical regression check, a full rubric review and the merge.

The remaining open work was not picked up, for a single reason: bringing an existing pull request to a mergeable state is treated as one complete unit of work per cycle. No judgement about the rest of the backlog is implied by its being untouched. Specifically:

- **#2001** is next in line. It remains human-gated on its own merits regardless of ordering, because `src/main/resources/config.yml` is modified and a changed default silently alters behaviour on every existing install. The open question recorded on it — that `factions.nonMembersCanInteractWithEntities: true` makes armour stands editable by non-members, which is weaker than `main` protects today — is a maintainer decision. It now needs a further rebase, since this merge landed ahead of it.
- **#1918** and **#1915** are drafts and were left as drafts.
- **#2002** was raised by this pull request's own analysis and is the correct home for the follow-up work; it is a design decision about four new listeners rather than a mechanical port, so it was not attempted here.
- **#2010** was filed this cycle from the in-game finding recorded above, so that it is not lost now that this pull request is closed.

This comment was posted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

PR #2001, opened 2026-08-08
Extend entity interaction protection beyond villagers and armour stands

### LINKED ISSUE #1996: Entity interaction protection only covers villagers and armour stands
## Summary

Territory protection for right-clicking entities is applied inconsistently. `PlayerInteractEntityListener` guards **villagers only**; every other entity type is left to `PlayerInteractAtEntityListener`, whose coverage depends on which of the two events the server fires for a given entity and version.

The result is that interactions with minecarts, boats, item frames, animals and similar entities inside another faction's claim may not be blocked at all. Chest and hopper minecarts and item frames make this a potential theft vector rather than a cosmetic gap.

## Origin

Raised on Discord alongside the eating-in-claims fix, phrased as a concern that holding food or a bow might allow armour stands, pigs or minecarts to be right-clicked.

That specific worry does not apply: entity interactions travel on `PlayerInteractEntityEvent` / `PlayerInteractAtEntityEvent`, not `PlayerInteractEvent`, so the edible-item bypass in `PlayerInteractListener` cannot leak into them. Checking the claim, however, surfaced the broader inconsistency described here, which is worth resolving on its own merits.

## Current behaviour

`src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractEntityListener.kt`:

```kotlin
val clickedEntityType = clickedEntity.type
if (clickedEntityType == EntityType.VILLAGER) {
    // ... claim + protectVillagerTrade flag check
}
```

No claim check is performed for any other entity type in this listener.

`src/main/kotlin/com/dansplugins/factionsystem/listener/PlayerInteractAtEntityListener.kt` does apply a general claim check and cancels when `isInteractionAllowed` returns false, which is what protects armour stands today.

Because `PlayerInteractAtEntityEvent` extends `PlayerInteractEntityEvent`, and because which event a server dispatches for a given entity has varied across Minecraft and Spigot versions, the effective protection for non-villager entities is version-dependent and untested.

## Suggested approach

1. **Confirm the gap in-game before changing code.** On a supported server version, from a player with no interaction rights in a claim, the following should be attempted inside that claim: opening a chest minecart, opening a hopper minecart, rotating and removing an item frame, entering a boat or minecart, leading and saddling animals, and using a lead on a fence.
2. For whichever of these are not blocked, extend `PlayerInteractEntityListener` to apply the same `claimService.isInteractionAllowed(...)` check the `AtEntity` listener already uses, keeping the existing `mf.bypass` handling and messaging.
3. The villager-specific path must be preserved as-is, since it is additionally gated on the `protectVillagerTrade` faction flag and should not be folded into a blanket check.
4. Consideration should be given to whether a general check needs its own faction flag or config toggle for parity with `nonMembersCanInteractWithDoors`, rather than being unconditional — servers relying on the current permissiveness would otherwise see a behaviour change on upgrade.

## Acceptance

- Entities inside a claim cannot be right-clicked by players without interaction rights, for whichever entity types step 1 identifies as unprotected.
- Villager trade protection continues to respect the `protectVillagerTrade` flag.
- `mf.bypass` continues to work, with the existing bypass message.
- Behaviour in wilderness continues to follow `wilderness.interaction.prevent`.
- Listener tests are added for the newly covered entity types.

---

dmccoystephenson:
Closes #1996

## What was wrong

`PlayerInteractEntityListener` performed a claim check for `EntityType.VILLAGER` only, additionally gated on the `protectVillagerTrade` faction flag. No other entity type was checked there. `PlayerInteractAtEntityListener` did apply a general `claimService.isInteractionAllowed(...)` check, and that is what protected armour stands.

Because `PlayerInteractAtEntityEvent` extends `PlayerInteractEntityEvent`, coverage for everything else depended on which of the two events a given server version raised for a given entity. Chest and hopper minecarts and item frames made that a theft vector rather than a cosmetic gap.

Two things were established from the source and the Bukkit API contract:

- The two events carry **separate handler lists**. `PlayerInteractAtEntityEvent` declares its own `getHandlerList()` (confirmed against the `spigot-api` 1.21.11 class file), so a single event object is delivered to exactly one of the two listeners. There was never a same-object double dispatch.
- A single physical right-click can nonetheless raise **both** events, because the client sends an "interact at" packet and follows it with a plain "interact" packet whenever the entity does not consume the first one. Cancelling the "interact at" event does not stop the second packet from arriving.

Both were subsequently confirmed on a live server — see the in-game verification comment on this pull request.

## What was changed

- The claim and wilderness check has been extracted into a new `EntityInteractionProtection`, and is now applied by both listeners. A single shared instance is constructed in `MedievalFactions` and handed to both.
- Since one right-click can raise both events, notifications are de-duplicated per player and clicked entity over a 250 ms window. Both events are still cancelled — each is a separate opportunity for the interaction to go through — but the player is told once. The retained per-player state is dropped on quit.
- A defensive `if (event is PlayerInteractAtEntityEvent) return` guard has been added to `PlayerInteractEntityListener`, so a single event could not be handled twice even if the handler lists were ever merged.
- The villager path has been preserved verbatim, moved into a private method. Villagers are deliberately **excluded** from the general check: folding them in would block trading in a claim whose faction has turned `protectVillagerTrade` off.
- Wilderness is handled exactly as the `AtEntity` listener already handled it, via `wilderness.interaction.prevent` and `wilderness.interaction.alert`.
- `mf.bypass` handling and all existing language keys are unchanged. No new language keys were needed — `CannotInteractWithEntityInFactionTerritory`, `CannotInteractWithEntityInWilderness` and `FactionTerritoryProtectionBypassed` are reused.
- The new config option `factions.nonMembersCanInteractWithEntities` has been documented in `config.yml` and in `CONFIG.md`, and a `CHANGELOG.md` entry has been added under `## [Unreleased]`.

## Verification

The behaviour has been observed on a live server. The full matrix, method and controls are in the in-game verification comment on this pull request; the summary is:

- On `main`, a non-member inside another faction's claim **opened** a chest minecart and a hopper minecart, **inserted an item into and rotated** an item frame, and **saddled** a pig — while simultaneously being told the interaction was denied. On this branch all four are prevented.
- Armour stands were already protected on `main` and remain protected here.
- In wilderness, and for a member of the owning faction, every entity stayed fully interactive on both builds with no messages, so the new check denies nothing it should allow.
- Exactly one protection message arrived per right-click across all in-claim rows, so the 250 ms de-duplication window is neither over- nor under-suppressing in the two-packet case.

Boat and minecart entry, villager trading (a summoned villager has no profession and so no trade GUI), leading an animal, `mf.bypass`, lead-to-fence, and wilderness with `prevent: true` were **not** decidable with that rig and remain unverified in game.

## Open question for the maintainer: the config default and its scope

`factions.nonMembersCanInteractWithEntities` has been named after the existing `factions.nonMembersCanInteractWithDoors` and defaulted to `false`, meaning **protection is on**. That was an assumption made here, on the grounds that it matches how MF's other claim protections default and that the issue describes the gap as a theft vector.

The other reading is defensible: servers relying on today's permissiveness will see a behaviour change on upgrade, and a default of `true` would make the upgrade silent and leave protection opt-in. **This is the maintainer's call.** Flipping it is a one-line change to `src/main/resources/config.yml` plus the `**Default:**` line in `CONFIG.md`; no code change is required, because the code reads the key with an implicit `false` fallback.

A related decision, now backed by the in-game evidence: the option gates the general check in **both** listeners, so setting it to `true` also stops armour stands being protected — which is *more* permissive than releases before this option existed. It is therefore not a restore-previous-behaviour switch. `CHANGELOG.md` and `CONFIG.md` have been corrected in this branch to say so plainly rather than to imply otherwise. Keeping the armour-stand path ungated regardless of the option would make `true` genuinely equal to today's behaviour, at the cost of making the option's scope depend on which event a server version raises. Either can be adopted before merge; no change has been made here, because the choice is a product decision rather than a defect.

One adjacent quirk was noticed and deliberately **not** changed, since it predates this work and is cosmetic: the `AtEntity` listener's general check also fires for villagers, so a player trading in a claim where `protectVillagerTrade` is off may still receive the "cannot interact with entities" notice even though the trade opens.

## Testing

- `./gradlew clean test` passes locally on the rebased head against JDK 17: **581 tests, 0 failures**, including the tests in `PlayerInteractEntityListenerTest`.
- `./gradlew lintKotlin` passes locally.
- CI is green on the head commit: `build`, `docker-build` and `changes` all pass.
- The tests cover chest minecart, hopper minecart, item frame, boat, minecart, animal and armour stand in a claim without interaction rights; the config toggle; a player with interaction rights; `mf.bypass`; wilderness with prevention on and off; both villager flag states; the `PlayerInteractAtEntityEvent` guard; and the both-events-for-one-click case asserting two cancellations and one message.

## Data safety

No schema migration. No DPC contract change. No lang-file key removed. One **added** `config.yml` key, `factions.nonMembersCanInteractWithEntities`, defaulting to `false` — an added default changes behaviour on existing installs, so this pull request is on the do-not-auto-merge list and is left open for the maintainer rather than merged by the loop.

## Backlog note for this cycle

Grooming this pull request was the whole of this cycle's work, one pull request per cycle, so the rest of the open backlog was deferred for that reason alone rather than on any judgement about the individual issues.

---


dmccoystephenson:
## In-game verification

The smoke test called for in the description has now been run on a live server, and the results are below. Nothing has been merged.

**Rig:** Spigot 1.21.1, MedievalFactions built from `fix/1996-entity-protection` and compared against `origin/main` (`abeab932`), driven by real bot clients so that genuine right-click packets are sent (`interact_at` followed by `interact`, the pair this PR de-duplicates). A single chunk was claimed by faction `Placeholderia`; the clicking player held no membership in it.

**How state was observed.** An earlier attempt at this matrix was discarded because it trusted client-side state, which reported mounts for cows and item frames. Every result below is instead read back from the server over RCON as an entity-NBT diff taken either side of the click, with naturally-drifting keys calibrated away by two pre-click snapshots. Container opens, which mutate no NBT, are observed as window opens. Each row is bracketed by an RCON round-trip, and any row where the server thread stalled is discarded rather than reported.

### Non-member, inside the claim

| Entity | Signal | `main` | This PR | This PR, toggle `true` |
| --- | --- | --- | --- | --- |
| Chest minecart | container opened | **opened** | not opened | opened |
| Hopper minecart | container opened | **opened** | not opened | opened |
| Item frame | `Item` / `ItemRotation` | **item inserted, frame rotated** | unchanged | item inserted, frame rotated |
| Pig | `Saddle` | **saddled** | unchanged | saddled |
| Armour stand | `ArmorItems` | unchanged | unchanged | **changed — see below** |

The theft vector described in #1996 reproduces on `main` exactly as characterised: the chest and hopper minecarts open, the item frame accepts an item and rotates, and another faction's pig is saddled — **while the player is simultaneously told "You cannot interact with entities in Placeholderia territory."** The notification was being sent by the `AtEntity` listener while the interaction itself went through on the second packet. On this PR all four are prevented.

Armour stands were already protected on `main`, and remain protected here, so the existing behaviour is preserved.

**Controls.** In wilderness (`wilderness.interaction.prevent: false`) and for a member of the owning faction, every entity above stayed fully interactive on both builds, with no messages. The new check therefore denies nothing it should be allowing.

### The de-duplication window holds

On this PR, in all nine in-claim rows, **exactly one** protection message was received per right-click, and no duplicate was seen in any run. This is meaningful rather than incidental: the interaction is now blocked, which is only possible if the plain-`interact` path is being cancelled as well — that path was not cancelled on `main` — and one message is nonetheless all that arrives. The 250 ms window is neither over- nor under-suppressing in the two-packet case.

### The config default, and the armour-stand consequence

The description raises two decisions. Evidence for the second is now available.

With `factions.nonMembersCanInteractWithEntities: true`, the previous permissiveness returns as intended. It goes **further than `main` did**, however: the armour stand, which `main` protects, becomes editable by a non-member. So the toggle is not a restore-previous-behaviour switch — setting it to `true` gives up protection that predates this PR.

That makes the description's own framing exact: gating both listeners on the option means the option's name promises less than it takes away. If the option is meant as an upgrade escape hatch, keeping the armour-stand path ungated regardless of it would leave `true` genuinely equal to today's behaviour rather than weaker than it. This is offered as evidence, not as a request for a particular outcome — the call remains the maintainer's.

### Not decidable with this rig

No claim is made either way about these, because the permissive controls did not exercise them either:

- **Boat and minecart entry** — no mount occurred in wilderness or as a member, so the bot never rides; a "not mounted" result says nothing about MF.
- **Villager trading** — a summoned villager has no profession and so no trade GUI to open. Both flag states of `protectVillagerTrade` are therefore still unverified in game. Worth noting that `You cannot trade with this villager.` was observed on both builds and in all three configurations, so that path is at least still reached.
- **Leading an animal** — the post-click read of the cow failed consistently in every scenario, including the permissive ones, so the measurement rather than the behaviour is what failed.
- **`mf.bypass`**, **lead-to-fence**, and **wilderness with `prevent: true`** were not exercised in this pass.

### Adjacent, and out of scope for this PR

The server was lost once during this work, unrelated to these changes: Spigot's watchdog terminated it after the main thread blocked for 60 s in `PlayerQuitListener.onPlayerQuit` → `MfPlayerService.save` → a synchronous jOOQ upsert. It is reproducible on player disconnect against the H2 backend and is present on `main` as well. It is raised here only so it is not lost; a separate issue would be the right home for it.

*Verified by Claude on behalf of Daniel Stephenson.*

---

dmccoystephenson:
## Status note from this cycle

This pull request was not groomed this cycle; #1999 was taken as the oldest of the three open pull requests from these cycles, rebased onto `main` and merged. One pull request is groomed per cycle.

Two mechanical notes for the next pass:

- The branch is behind `main` by the JSON storage backend, the dev-release workflow, the gate-distance fix and the merge of #1999, so a rebase is needed before it can be assessed.
- No check is reported here. That is not a failure: the head commit was pushed by the Copilot agent while addressing review comments, so the required `Build` workflow is held in `action_required` and the pull request reads as `BLOCKED`. A rebase and force-push from a human account re-triggers it.

Beyond that, this pull request will not be merged autonomously in any cycle. `src/main/resources/config.yml` is modified, and a changed or added default silently alters behaviour on every existing install, which the data-safety charter this loop follows treats as human-gated. The open question raised in the verification comment above stands on its own merits as well: with `factions.nonMembersCanInteractWithEntities: true`, armour stands become editable by non-members, which is weaker than the protection `main` provides today, so the option as written gives up more than its name promises. That call belongs to the maintainer.

This comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

dmccoystephenson:
## Grooming and self-review

This pull request has been brought current with `main` this cycle. It had gone `CONFLICTING` behind thirty commits, and no check was being reported because its previous head was pushed by the Copilot agent, which held the required `Build` workflow in `action_required`. It has been rebased onto `fa154e16`, the one conflict (`CHANGELOG.md`, both sides kept) resolved, and force-pushed, which re-triggered CI. Head is now `06cde7fe`.

Two documentation lines were corrected in the same push, because the in-game verification above had made them inaccurate: `CHANGELOG.md` and `CONFIG.md` both implied that `factions.nonMembersCanInteractWithEntities: true` restores the behaviour of earlier releases, when in fact it is more permissive than they were — armour stands were protected before this option existed, and `true` leaves them editable. No code was changed; the choice of whether to keep the armour-stand path ungated remains open and is the maintainer's.

### Rubric

- **Scope:** PASS — nine files, each traceable to #1996: the extracted `EntityInteractionProtection`, the two listeners that now share it, `PlayerQuitListener` for the retained-state cleanup, `MedievalFactions` for the shared instance, the config key, its two documentation entries, and the tests. No unrelated formatting or renames appear in `git diff origin/main...HEAD`.
- **Tests-new:** PASS — both public members of `EntityInteractionProtection` are exercised: `protect` throughout the seventeen tests, and `forgetPlayer` by `onPlayerInteractEntity_AfterPlayerIsForgotten_ShouldNotifyAgainWithinWindow`.
- **Tests-fix:** PASS — verified empirically rather than by reasoning. With the single call to `entityInteractionProtection.protect(...)` neutralized in `PlayerInteractEntityListener`, `./gradlew test --tests '*PlayerInteractEntityListenerTest*'` reports **11 of 17 failed**, covering chest minecart, hopper minecart, item frame, boat, minecart, animal, armour stand, wilderness, bypass, the both-events case and the forget-player case. With the call restored the same run is green. The regression coverage is therefore real and not a false negative.
- **Migrations-additive:** PASS — `git diff --name-only origin/main...HEAD -- '*db/migration/*'` returns nothing. No migration is added, edited or removed.
- **Lang-externalized:** PASS — every `sendMessage` added by the diff resolves its text through `plugin.language[...]`; the three keys used (`CannotInteractWithEntityInFactionTerritory`, `CannotInteractWithEntityInWilderness`, `FactionTerritoryProtectionBypassed`) already existed, and no lang file is modified.
- **DPC-contract:** PASS — nothing under `dpc/` or `pacts/` is touched.
- **Main-thread-safety:** PASS — both handlers are ordinary synchronous Bukkit event handlers, and no asynchronous dispatch is introduced. The claim lookup they perform is the same one `PlayerInteractAtEntityListener` already performed on the same thread, so the threading posture is unchanged.
- **Plugin.yml:** PASS, not applicable — no command or permission is added.
- **Sibling structure:** PASS — `EntityInteractionProtection` sits alongside the other classes in `listener/`, and the new test mirrors the structure of its neighbours in `src/test/kotlin/.../listener/`.
- **Docs:** PASS — `CONFIG.md` documents the new key with type, default and description, and `CHANGELOG.md` carries an entry under `## [Unreleased] / ### Fixed`. `USER_GUIDE.md`, `README.md` and `README.pt-br.md` make no statement about entity interaction and so need no change.
- **Issue resolution:** PASS — #1996 names `PlayerInteractEntityListener`'s villager-only check as the defect, and that is the code this pull request changes.
- **CI:** PASS — `build`, `docker-build` and `changes` are all green on `06cde7fe`. Locally, `./gradlew clean test` passes at 581 tests with 0 failures and `./gradlew lintKotlin` is clean, on JDK 17.
- **Config-defaults:** **FAIL, by design** — `factions.nonMembersCanInteractWithEntities` is added to `src/main/resources/config.yml` with a default of `false`, which turns protection on for every existing install on upgrade. That is exactly the kind of silent behaviour change the data-safety charter treats as human-gated, so this pull request will not be merged by the loop under any circumstances. It is left open, green and review-ready for the maintainer.

### Out-of-diff observations

- `## [Unreleased]` in `CHANGELOG.md` currently carries **two** separate `### Fixed` sections, one before `### Added` and one after `### Changed`. This predates this branch and arrived on `main` independently; it has deliberately not been fixed here, since moving the intervening entries would bury this pull request's real change in unrelated churn. It is worth a small tidying pull request of its own.
- The default constructor argument on both listeners, `entityInteractionProtection: EntityInteractionProtection = EntityInteractionProtection(plugin)`, exists for the tests. It is worth noting that if either listener were ever registered without the shared instance being passed, de-duplication would silently stop working, because the two listeners would hold separate caches. `MedievalFactions` does pass the shared instance to all three registrations today.

### Remaining decision

Nothing further is proposed here. The two questions in the description — the default value, and whether the option should gate the armour-stand path at all — are product decisions and are left to the maintainer.

---

Issue #2006, opened 2026-08-09
Several commands read the sender's Location off the main thread inside async tasks


dmccoystephenson:
## Finding

Several command classes read the sender's `Location` — and dereference `Chunk`, `World` and `Block` from it — from inside the `Runnable` passed to `plugin.server.scheduler.runTaskAsynchronously`, rather than snapshotting it on the main thread before the task is dispatched.

Two commands already demonstrate the correct pattern, which makes the inconsistency easy to see:

`MfFactionSetHomeCommand` (line 26) snapshots first, then dispatches:

```kotlin
val senderLocation = sender.location
plugin.server.scheduler.runTaskAsynchronously(
    plugin,
    Runnable {
```

`MfFactionMapCommand` (line 46) does the same with `sender.location.chunk`.

The following read from inside the async task instead:

- `MfFactionClaimCheckCommand` line 26 — `claimService.getClaim(sender.location.chunk)`
- `MfFactionClaimCircleCommand` lines 70–71 — `sender.location.chunk.x` / `.z`
- `MfFactionClaimFillCommand` lines 65–67 — `sender.location.world?.uid`, `sender.location.chunk.x` / `.z`
- `MfFactionUnclaimCommand` line 62 — `val senderChunk = sender.location.chunk`
- `MfGateRemoveCommand` line 54 — `MfBlockPosition.fromBukkitLocation(sender.location)`, which additionally calls `World.getBlockAt(Location)` inside `MfBlockPosition.fromBukkitLocation`
- `MfDuelAcceptCommand` lines 94 and 128–131 — `MfPosition.fromBukkitLocation(sender.location)` and `it.location.distanceSquared(...)` over `server.onlinePlayers`

## Impact

Bukkit's API is not thread-safe and is documented as main-thread only. Off-thread reads of entity position are torn rather than merely stale: `Location` is mutated in place on the server thread, so a coordinate read off-thread can mix pre-move and post-move components. `sender.location.chunk` and `World.getBlockAt(...)` go further and can trigger chunk lookup or load from an async thread.

In practice this surfaces as a player who moves during the command dispatch acting on the wrong chunk — claiming, unclaiming, or removing a gate one chunk away from where they intended — and, less often, as an async-access exception from Paper's thread checks. Because `/faction unclaim` and `/faction gate remove` destroy records, a torn read there costs data rather than just producing a confusing message.

[...]

PR #2007, opened 2026-08-09
fix: make gate distance world-aware and overflow-free

### LINKED ISSUE #1991: MfCuboidArea.distanceSquared overflows Int, letting gates be removed from far beyond gates.maxRemoveDistance

### LINKED ISSUE #1993: MfCuboidArea.distanceSquared ignores worlds, so /f gate remove can delete a gate in another world
## Finding

`MfCuboidArea.distanceSquared` compares coordinates only; it never compares `worldId`. Its sibling `contains` does:

```kotlin
fun contains(position: MfBlockPosition) =
    position.worldId == minPosition.worldId &&
        position.x >= minPosition.x &&
        ...

fun distanceSquared(position: MfBlockPosition): Int {
    val x = position.x
    val y = position.y
    val z = position.z
    ...
}
```

So an area in one world reports a distance of `0` for a position with the same coordinates in a different world.

## Impact

`MfGateRemoveCommand` selects the gate to delete purely by this distance, over every gate belonging to the sender's faction:

```kotlin
val existingGates = gateService.getGatesByFaction(faction.id)
val gateDistanceSquared = existingGates.associateWith { gate ->
    MfBlockPosition.fromBukkitLocation(sender.location)?.let { gate.area.distanceSquared(it) }
}
val closestGate = gateDistanceSquared.keys.minByOrNull { gateDistanceSquared[it] ?: Int.MAX_VALUE }
```

`MfGateService.getGatesByFaction` applies no world filter:

```kotlin
fun getGatesByFaction(factionId: MfFactionId) = gatesById.values.filter { it.factionId == factionId }
```

A faction with gates in more than one world (Overworld and Nether, for instance) is therefore exposed to the wrong gate being deleted: a player standing near coordinates in the Overworld that happen to match a gate in the Nether gets that Nether gate returned as the "closest" one, at a reported distance of `0`, and `/f gate remove` deletes it. Nether coordinate compression makes overlapping x/z between the two worlds routine rather than exotic.

Because gate deletion is not confirmed and not undoable, this destroys a build's gate record without the player having any indication that the gate they were standing next to was not the one removed.

## Current behaviour is characterized by a test

`MfCuboidAreaTest.testDistanceSquaredIgnoresTheWorldOfTheGivenPosition` asserts the world-blind result so the behaviour is pinned rather than endorsed. That test must be updated when this is fixed.

## Suggested fix

Either `distanceSquared` would return `null` (or `Int.MAX_VALUE`) for a position in a different world, matching the world-awareness of `contains`, or `MfGateRemoveCommand` would filter `existingGates` to the sender's world before measuring. The former is preferable: it fixes every future caller rather than one, and it makes the two methods on `MfCuboidArea` consistent with each other.

Related: #1991 covers a separate defect in the same method (`Int` overflow at long distances). Both affect the same gate-removal selection and could reasonably be fixed together.

---

dmccoystephenson:
## Summary

Two defects in `MfCuboidArea.distanceSquared` are fixed, both of which land on `/faction gate remove` — the method's only production caller — and both of which destroy a gate record that was not the one the player intended to remove.

- **Worlds were ignored.** Coordinates alone were compared, so an area in one world reported a distance of `0` for the same coordinates in another world, while its sibling `contains` has always compared `worldId`. A faction holding gates in both the Overworld and the Nether — where overlapping x/z is made routine by coordinate compression — could have the wrong gate silently deleted, with no confirmation prompt and no way to undo it. `distanceSquared` now returns `null` for a position in another world, so the two methods on `MfCuboidArea` agree with each other and every future caller inherits the fix rather than one call site.
- **`Int` arithmetic overflowed.** The squared distance was computed in `Int`, which wraps past `Int.MAX_VALUE` from roughly 46,341 blocks away on a single axis and can turn negative. A negative value passed the `distanceSquared > maxRemovalDistanceSquared` guard, so `gates.maxRemoveDistance` was defeated, and it also made a very distant gate compare as the nearest one under `minByOrNull`. The computation and the return type are now `Long`.

`MfGateRemoveCommand` was widened to match: the `minByOrNull` fallback became `Long.MAX_VALUE`, so a gate in another world sorts last and is never chosen while a same-world gate exists, and the case where *every* gate is in another world is caught by the existing `distanceSquared == null` guard. The misleading local name `minRemovalDistanceSquared` was corrected to `maxRemovalDistanceSquared`, since it is an upper bound.

Closes #1991
Closes #1993

## What was changed

- `MfCuboidArea.distanceSquared` — returns `Long?`, short-circuits on a world mismatch, and computes the three squared deltas in `Long`. The contract for `null` is documented in KDoc so a caller cannot mistake it for a distance.
- `MfGateRemoveCommand` — `Long` throughout; `maxRemoveDistance` is widened before squaring so the config value cannot overflow either.
- `MfCuboidAreaTest` — the two characterization tests that pinned the old behaviour (`testDistanceSquaredIgnoresTheWorldOfTheGivenPosition` and `testDistanceSquaredOverflowsForVeryDistantPositions`, both added by #1992 with an explicit note that they were to be updated when this was fixed) now assert the corrected behaviour. The remaining geometry assertions were widened to `Long` literals.
- `MfGateRemoveCommandTest` — new, following the fixture and scheduler-capture pattern established by `MfFactionPowerCommandTest`. It covers the command level, which is where the data loss actually occurs.
- `CHANGELOG.md` — two entries under `## [Unreleased]` → `### Fixed`.
- `COMMANDS.md` and `USER_GUIDE.md` — both described a click-then-confirm removal flow that has never existed. `/gate remove` deletes the faction's nearest gate as soon as it is run, with no click and no confirmation prompt. The descriptions now say so, and both note that gates in other worlds are not candidates. This drift predates the branch and is wider than the two issues; it was included because it describes the very selection behaviour this change alters.

## Data safety

No schema migrations; no DPC contract change; no `config.yml` default change; no `plugin.yml` change; no lang-file change; no new user-facing string. Nothing persisted or on a wire format is touched — the change is confined to an in-memory geometry calculation and the gate-selection logic that reads it. The effect on live-server data is strictly protective: gate records that could previously be destroyed by mistake are no longer reachable by `/faction gate remove`.

The change is behaviour-narrowing rather than behaviour-widening, which is worth naming explicitly for review: a player who was previously able to remove a gate at extreme range, or across worlds, no longer can. That is the intended fix, but it is a change to what the command will accept.

## Test plan

- [x] `MfCuboidAreaTest` — cross-world positions return `null`; a 60,000-block offset returns `3,600,000,000` rather than the previously-pinned `-694967296`; same-world face, corner and containment distances are unchanged apart from widening to `Long`.
- [x] `MfGateRemoveCommandTest` — the nearest of two same-world gates is the one deleted; a gate at the sender's exact coordinates in another world is *not* deleted in favour of the same-world gate, even when the other-world gate is measured first; a lone other-world gate produces `CommandGateRemoveFailedToFindGate` and no deletion; a lone gate 60,000 blocks away likewise produces no deletion.
- [ ] Manual smoke test on a server with a faction holding gates in two worlds — not performed here, and recommended before release.

The new tests were confirmed to be genuine regression tests rather than argued to be. A temporary commit restored the pre-fix semantics — world-blind comparison and `Int`-width arithmetic — while keeping the new `Long?` signature so the suite still compiled, isolating behaviour from types. CI reported `401 tests completed, 5 failed`: both updated `MfCuboidAreaTest` cases and all three `MfGateRemoveCommandTest` protection cases. `testOnCommand_removesTheGateClosestToTheSender` passed under the reverted code, as it should. The commit was then dropped and the branch force-pushed back.

That experiment also exposed a weakness in the tests themselves, since fixed: the protection cases failed with an exception from an unstubbed `MfGateService.delete` rather than with the `never()` verification they were written around. Deletion is now stubbed for every gate in each fixture, including the ones that must be left alone, so a future regression points at the assertion that describes the protection.

## Verification anchor

The local Gradle wrapper cannot run in this environment, so the CI check on this pull request's head SHA is the anchor rather than a local `./gradlew clean test`. Every changed file is Kotlin source or a resource inside the Gradle build, so CI's scope covers all of them; the result is reported in the checks on this pull request.

## Deferred backlog

The rest of the open issue backlog was left untouched: this cycle was scoped to the pair of `MfCuboidArea` defects, which share a method and a call site and were noted on #1993 as reasonably fixed together. One adjacent observation is recorded rather than fixed here, so it is not lost — `MfGateRemoveCommand` reads `sender.location` from inside an async task, which is Bukkit API access off the main thread. It is pre-existing and out of scope for this change, and is filed separately as #2006, which found the same pattern in five other command classes.

---


dmccoystephenson:
## Self-review

Posted as a plain comment rather than a formal review object, so that no reviewer independence is implied.

**Anchor:** CI is green on head `6bb8151b` — `build` pass (3m16s), `docker-build` pass, `changes` pass. `:test` executed rather than being reported up-to-date; 401 tests were collected on the run used for the regression check below.

### Rubric

- **Scope: PASS, with a declared widening.** Five of the seven changed files are the two production files, their tests, and `CHANGELOG.md`. The sixth and seventh (`COMMANDS.md`, `USER_GUIDE.md`) correct drift that predates this branch: both described a click-then-confirm removal flow that `MfGateRemoveCommand` has never implemented. It was included rather than deferred because the drift is about *which gate `/gate remove` deletes*, which is the exact behaviour this change alters, and because a reader following the old text would not understand why a wrong-gate bug was possible at all. It is nonetheless wider than `Closes #1991` / `Closes #1993`, and is called out here rather than left for a reviewer to notice.
- **Tests-new: PASS.** `distanceSquared` is the only changed public method. It is covered directly in `MfCuboidAreaTest` and through its single production caller in the new `MfGateRemoveCommandTest`.
- **Tests-fix: PASS — proven empirically, not by reasoning.** A temporary commit restored the pre-fix semantics (world-blind comparison, `Int`-width arithmetic) while keeping the new `Long?` signature so that the suite still compiled, isolating behaviour from types. CI run `31341624998` reported `401 tests completed, 5 failed`: `MfCuboidAreaTest.testDistanceSquaredDoesNotOverflowForVeryDistantPositions`, `MfCuboidAreaTest.testDistanceSquaredIsNullForAPositionInAnotherWorld`, and all three `MfGateRemoveCommandTest` protection tests. `testOnCommand_removesTheGateClosestToTheSender` passed under the reverted code, as it should — it does not depend on either defect. The commit was then dropped and the branch force-pushed back.
- **Test-failure legibility: initially FAIL, fixed in `766f531f`.** That experiment exposed a real weakness: the three protection tests failed with `kotlin.NoWhenBranchMatchedException` raised by an unstubbed `MfGateService.delete`, not by the `never()` verification they were written around. A future maintainer would have been handed a confusing stack trace instead of the assertion describing the protection. Deletion is now stubbed for every gate in each fixture, including the ones that must be left alone.
- **Sibling structure: PASS.** `MfGateRemoveCommandTest` follows `MfFactionPowerCommandTest` — same `TestUtils.CommandTestFixture`, the same `mockServices` / `mockLanguageSystem` / `mockConfig` / `mockScheduler` / `mockLogger` helper split, and the same scheduler stub that runs the dispatched `Runnable` inline so the command body is genuinely exercised.
- **Sibling renames: PASS.** `minRemovalDistanceSquared` became `maxRemovalDistanceSquared`, since it is an upper bound; it has no parallel sibling elsewhere.
- **Docs: PASS.** `CHANGELOG.md` gains two `### Fixed` entries under `## [Unreleased]`. `CONFIG.md`'s `gates.maxRemoveDistance` entry was checked and is still accurate. No command, config option or faction flag was added, so `FACTION_FLAGS.md`, `DATABASE_QUERYING.md`, `DEPENDENCY_USAGE.md` and the two READMEs are untouched by design.
- **Issue resolution: PASS.** #1991 named `Int` overflow in `distanceSquared` and the `Int.MAX_VALUE` fallback in `MfGateRemoveCommand`; both are changed. #1993 named the missing world comparison and preferred fixing it in `distanceSquared` over filtering in the command; that is what was done. The two characterization tests added by #1992, which both issues flagged as needing an update, were updated rather than deleted.
- **Migrations-additive: PASS, vacuously.** `git diff origin/main...HEAD -- '*db/migration/*'` is empty.
- **Lang-externalized: PASS.** No string literal was added. The rejection path reuses the existing `CommandGateRemoveFailedToFindGate` key.
- **DPC-contract: PASS, vacuously.** Nothing under `dpc/` or `pacts/` is touched.
- **Main-thread-safety: FAIL — pre-existing, not introduced, and deferred.** `MfGateRemoveCommand.kt:54` reads `sender.location` from inside the `runTaskAsynchronously` body, and `MfBlockPosition.fromBukkitLocation` then calls `World.getBlockAt`. This line was edited by this change and was left off-thread. It is scored FAIL rather than not-applicable because the rule is genuinely violated on a touched line. It was deferred rather than fixed because the same pattern occurs in five other command classes, which makes it a coherent separate change; it is filed as #2006 with all six sites listed and with `MfFactionSetHomeCommand` / `MfFactionMapCommand` cited as the correct pattern.
- **Plugin.yml: PASS, vacuously.** No command or permission was added.
- **Config-defaults: PASS.** `src/main/resources/config.yml` is untouched; `gates.maxRemoveDistance` is only read, and is now widened to `Long` before being squared so the bound itself cannot overflow.

### Further findings, folded in rather than anchored inline

- `src/main/kotlin/com/dansplugins/factionsystem/command/gate/remove/MfGateRemoveCommand.kt:53-55` — `MfBlockPosition.fromBukkitLocation(sender.location)` is evaluated once per gate inside `associateWith`, so a faction at the five-gate default performs five Bukkit reads where one snapshot would do, and the gates are compared against positions sampled at slightly different moments. Hoisting it would be a one-line improvement. It was left alone deliberately: #2006 has to restructure exactly this expression to move it onto the main thread, and doing half of that here would create a conflict for no behavioural gain today.
- `src/main/kotlin/com/dansplugins/factionsystem/area/MfCuboidArea.kt:75` — the JVM signature of a public method on a public class changed from `int distanceSquared(...)` to `Long distanceSquared(...)`. Nothing else in this repository calls it, but anything compiled against a previous Medieval Factions jar would need recompiling. No source-level break exists for a Kotlin caller beyond handling the new `null`, which is the point of the change.
- `src/main/kotlin/com/dansplugins/factionsystem/command/gate/remove/MfGateRemoveCommand.kt:56` — the `INFO` log line now prints `null` for every gate in another world. The line is pre-existing and logs on every invocation of the command at `INFO`, which is arguably too loud for a per-command diagnostic, but changing its level is out of scope here.
- The change narrows what `/gate remove` accepts: removal across worlds, and removal from beyond roughly 46,341 blocks, both stop working. That is the fix rather than a side effect, but it is a behaviour removal on live servers and is named here so it is reviewed as one.

### Data safety

No schema migration, no DPC payload or contract change, no `config.yml` default change, no `plugin.yml` change, no lang-file change. Nothing persisted is written differently. The effect on live-server data is protective: gate records that `/gate remove` could previously destroy by mistake are no longer reachable.

### Deferred backlog

The remainder of the open issue backlog was left untouched, for the single reason that this cycle was scoped to the paired `MfCuboidArea` defects — #1993 recorded that the two share a method and a call site and could reasonably be fixed together. #2006 was filed during the cycle rather than fixed.

---

This review comment was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

PR #2019, opened 2026-08-17
Snapshot the sender's position on the main thread in the three commands left by #2006

### LINKED ISSUE #2006: Several commands read the sender's Location off the main thread inside async tasks

dmccoystephenson:
Closes #2006

## What was wrong

Three command classes remained from the inventory in #2006; the other three were addressed in #2012 and are already on `main`. In each of the three, the sender's `Location` was read from inside the `Runnable` handed to `runTaskAsynchronously` rather than being snapshotted on the main thread first. Two consequences follow, and only the first is visible to a player:

- **The position is taken at task time, not at command time.** A player who kept walking had the command applied wherever they had arrived by then — a circle claimed around the wrong chunk, a flood fill started from the wrong chunk, or a duel recording a return position the duellist had already left.
- **The read itself is unsafe.** A `Location` is mutated in place on the server thread, so coordinates read from another thread can mix pre-move and post-move components, and `Location.getChunk()` can reach into chunk loading from off-thread.

Two corrections to that issue's inventory were established while reading these three, and have been acted on here rather than taken at face value:

- `MfDuelAcceptCommand` lines 128–131 are **not** an off-thread read. They sit inside a `runTaskAsynchronously` → `runTask` hop and therefore already execute on the main thread. Only the `MfPosition.fromBukkitLocation` pair was genuinely off-thread there. Those lines are left alone.
- The issue does not mention `sender.world`, which was read off-thread by both claim commands to decide whether claiming is blocked in that world. That read belongs to the same fix and is included.

## What was changed

- `MfFactionClaimCircleCommand` takes the sender's world and chunk coordinates in `onCommand`, before dispatching, and the inner main-thread task uses the snapshotted world rather than re-reading it.
- `MfFactionClaimFillCommand` does the same. The null-world check is deliberately left where it was, inside the asynchronous task, so the order in which its refusal messages are produced is unchanged.
- `MfDuelAcceptCommand` converts both players' locations to `MfPosition` in `onCommand`, and the saved duel carries those values.

No user-facing message, permission, command registration, configuration option, faction flag or language key is touched.

## What was deliberately left out

- `MfDuelAcceptCommand` also reads `sender.health` and `target.health` off-thread, two lines above the position reads. That is the same class of access, but it is not a `Location` read, it is not named by #2006, and changing when health is sampled changes what a duel restores — a design decision rather than a bug fix. It has been filed as #2021 for a maintainer decision rather than folded in here.
- `MfFactionMapCommand` reads `sender.world` from inside its asynchronous task, at line 65, even though it already snapshots the sender's chunk correctly on the main thread. This was noticed while surveying the command package and is not covered by #2006's inventory; it has been filed as #2020 rather than adding a fourth command to this changeset.
- The rest of the open backlog is deferred this cycle for the ordinary reason that one unit of work is taken per cycle. Pull request #2001 stays open and untouched: it is blocked on a maintainer decision about the default of a configuration option it adds, and cannot be advanced without one.

## Test plan

- [x] Three regression tests added, one per command, each holding the queued task, moving the sender to a position that would produce a different and separately-stubbed outcome, and only then running the task.
  - `MfFactionClaimCircleCommandTest` — two unclaimed chunks, saving either one stubbed; only the one stood on is claimed.
  - `MfFactionClaimFillCommandTest` — the chunk stood on is held by the sender's own faction, so the fill terminates there and reports nothing to claim; the chunk walked into is unclaimed and would report a lack of power instead, so which message arrives says which chunk the fill started from.
  - `MfDuelAcceptCommandTest` — both duellists are moved several hundred blocks after the command is run; the saved duel is asserted to carry the positions they held when it was run.
- [x] Supporting coverage added for all three commands, none of which previously had a test class: permission refusal, non-player sender, missing duel target, and the ordinary success path.
- [x] The three regression tests were empirically confirmed to fail without the production fix. A temporary commit reverting only the three production files was pushed to this branch, and the CI `build` check on it failed with exactly three failures — [run 32001482076](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/32001482076/job/95302542490), reporting `584 tests completed, 3 failed`, those three being `MfFactionClaimCircleCommandTest.testOnCommand_claimsTheChunkTheSenderStoodInWhenTheCommandWasRun`, `MfFactionClaimFillCommandTest.testOnCommand_fillsFromTheChunkTheSenderStoodInWhenTheCommandWasRun` and `MfDuelAcceptCommandTest.testOnCommand_recordsThePositionsHeldWhenTheCommandWasRun`. The temporary commit was then dropped from the branch and is not part of this changeset.
- [x] The CI `build` check is green on this pull request's head commit. **The local build was not usable in the environment this was written in** — Gradle 7.6.4 cannot parse the class file version of the installed JDK, so `./gradlew` fails during configuration and neither `test` nor `lintKotlin` was run locally. The CI `build` check on the head commit therefore stands as the anchor for this changeset in place of a local run.
- [ ] Not verified in game; no Minecraft server was available. A maintainer running `/faction claim circle`, `/faction claim fill` and `/duel accept` on a live server while moving is worth doing before release.

## Data safety

No schema migration. No DPC contract change. No `config.yml` default added or changed. No `plugin.yml` change. No language file touched, so no translation key is removed. The changed behaviour is confined to *when* a position is read; the claim and duel records written are the same records that were written before, keyed on the position the player actually occupied when the command was issued.

---


dmccoystephenson:
## Review

The full diff was read against `origin/main`, the three production changes were traced through their surrounding thread hops, and the claims made in the description were checked against source rather than taken at face value. Findings are folded into this comment as `path:line — finding` notes, since the review is being posted as a comment rather than as a formal review object.

### Rubric

- **Scope: PASS** — seven files. Three production files, each named in #2006's inventory; the three test classes mirroring them, none of which existed before; and one `CHANGELOG.md` line. Thirty net lines of production change against 746 of test. No unrelated formatting, rename or comment churn appears in the diff.
- **Tests-new: PASS** — no new public method is introduced, but all three changed `onCommand` bodies went from no test class at all to one each, covering the permission refusal, the non-player sender, the ordinary success path, and the regression itself.
- **Tests-fix: PASS — empirically, not by reasoning.** A temporary commit reverting only the three production files was pushed to this branch and CI was allowed to run on it: [run 32001482076](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/32001482076/job/95302542490) reported `584 tests completed, 3 failed`, and the three were exactly the three new regression tests. No other test was disturbed by the revert, so each regression test is pinned to its own fix rather than to incidental setup. The temporary commit has since been dropped from the branch, and the head now carries only the fix commit.
- **Sibling structure: PASS** — all three test classes use `TestUtils.CommandTestFixture`, `@TestInstance(PER_CLASS)`, and the `// prepare` / `// execute` / `// verify` block convention that neighbours such as `MfFactionClaimCheckCommandTest` and `MfFactionPowerCommandTest` already use.
- **Sibling renames: not applicable** — no identifier is renamed.
- **Docs: PASS** — a `CHANGELOG.md` entry is added under `## [Unreleased]`, in the same `### Fixed` block that carries the sibling async fixes from #2012. `COMMANDS.md`, `CONFIG.md`, `FACTION_FLAGS.md`, `USER_GUIDE.md`, `DATABASE_QUERYING.md`, `DEPENDENCY_USAGE.md` and both READMEs were each checked and need no change: no command, permission, option, flag, dependency or schema surface is altered, and the existing prose already describes the behaviour this change restores.
- **Issue resolution: PASS** — #2006 names six commands. Three (`MfFactionClaimCheckCommand`, `MfFactionUnclaimCommand`, `MfGateRemoveCommand`) were fixed in #2012 and are on `main`; the three here complete the list. The one part of the inventory this changeset does not touch was verified rather than assumed: `MfDuelAcceptCommand.kt:129-134`, the `it.location.distanceSquared(...)` reads over `world.players`, sit inside a `plugin.server.scheduler.runTask(...)` block nested in the async task, so they already run on the main thread and #2006's listing of them is mistaken. Closing #2006 is therefore correct rather than premature.
- **CI: PASS** — `build` green in 3m29s on head `28b34725`, run [32215788058](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/32215788058). The local Gradle wrapper is unusable in the environment this was written in (Gradle 7.6.4 against a newer JDK's class file version), so this CI check on the exact head commit is the anchor in place of a local `test`/`lintKotlin` run, and is stated as such in the description.
- **Migrations-additive: PASS, vacuously** — `git diff origin/main...HEAD -- '*db/migration/*'` is empty. No migration is added and none is edited.
- **Lang-externalized: PASS** — no string literal intended for a player is added. Every message the changed code produces goes through an existing `plugin.language[...]` key, and no `lang/` file is touched, so no translation is added, removed or re-encoded.
- **DPC-contract: PASS, vacuously** — nothing under `dpc/` or `pacts/` appears in the diff.
- **Main-thread-safety: PASS** — this is the substance of the change, and it was checked in the reverse direction as well, for newly introduced off-thread access rather than only for removed access. `senderWorld.getChunkAt(x, z)` in `MfFactionClaimCircleCommand` remains inside the inner `runTask` block, so moving the world reference earlier did not drag a chunk load off the main thread; every other snapshotted value crossing into an async task is a plain `Int`, `UUID` or `MfPosition`.
- **Plugin.yml: PASS, vacuously** — no command or permission is added.
- **Config-defaults: PASS, vacuously** — `config.yml` is not in the diff.

### Notes

- `src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt:39` — the world is captured twice over, from two separate reads: `senderWorld` comes from `sender.world` on line 39, while `senderWorldId` is derived from `senderLocation.world?.uid` on line 41. Both reads happen on the main thread within the same tick, so they cannot disagree and this is not a defect. It is worth noting only because the whole point of the change is that a position should be sampled once; taking the world from `senderLocation.world` (or `senderChunk.world`) would make that single-sample invariant visible in the code rather than dependent on the reader knowing that nothing can intervene between the two lines. The same observation applies to `MfFactionClaimCircleCommand.kt:34-35`.
- `src/main/kotlin/com/dansplugins/factionsystem/command/faction/claim/MfFactionClaimFillCommand.kt:41` — `senderLocation.chunk` is now evaluated eagerly, ahead of the `senderWorldId == null` guard that is deliberately kept inside the async task. Since `Location.getChunk()` resolves through the location's world, it is the call that would fail first if the world were ever null, and the guard against that is downstream of it. This is not a regression: the previous code evaluated `sender.location.chunk` unconditionally in exactly the same order, merely on the other thread. Keeping the guard where it is preserves the order in which the refusal messages are produced, which is what the description says was intended, so no change is asked for here.
- Two off-thread accesses adjacent to this work were found and filed rather than folded in, keeping the changeset to what #2006 names: #2021, for the `sender.health` / `target.health` reads at `MfDuelAcceptCommand.kt:92-93`, which are held back because moving them changes which health a duel restores and that is a maintainer's design decision rather than a bug fix; and #2020, for `sender.world` at `MfFactionMapCommand.kt:65`, which is read inside an async task even though that command already snapshots its chunk coordinates correctly.

### Data safety

No schema migration is added or edited, no DPC wire change is made, no `config.yml` default is altered, no `plugin.yml` registration changes, and no language file is touched. None of the paths on the do-not-auto-merge list appear in `git diff --name-only origin/main...HEAD`. The records written are the same records as before; only the moment at which the position feeding them is sampled has moved earlier, which is what makes them correct.

### Verdict

Ready to merge. The one gap left is that no in-game verification was possible, since no Minecraft server was available here — running `/faction claim circle`, `/faction claim fill` and `/duel accept` while moving is worth doing before release, and is recorded unchecked in the description's test plan.

---

Issue #2021, opened 2026-08-19
MfDuelAcceptCommand reads both players' health off the main thread inside its async task


dmccoystephenson:
## Summary

`MfDuelAcceptCommand` reads `target.health` and `sender.health` from inside the `Runnable` handed to `runTaskAsynchronously`, at `src/main/kotlin/com/dansplugins/factionsystem/command/duel/accept/MfDuelAcceptCommand.kt:92-93`:

```kotlin
challengerHealth = target.health,
challengedHealth = sender.health,
```

These two values are stored on the saved `MfDuel` and are what the duel restores each combatant to when it ends.

## Why it matters

This is the same class of access as #2006, applied to health rather than position:

- **The read is off the main thread.** `Player.getHealth()` is a live read of entity state that the server thread mutates; it is not documented as thread-safe.
- **The value is sampled at task time, not at command time.** Everything the task does before reaching this point is database work — an invite lookup, two player lookups and an invite deletion. A combatant who takes damage while that work is in flight has the post-damage health recorded as the health to restore.

## Why it was not fixed alongside the position reads

PR #2019 snapshots the two `Location` reads in this same constructor call on the main thread, and deliberately leaves these two health reads alone. The reasoning recorded there is that fixing them is not a pure timing correction the way the position fix is: moving the health read earlier changes *which* health a duel restores, and whether the intended value is the health held when `/duel accept` was typed or the health held when the duel record is written is a design decision rather than a bug fix. #2006 does not name these reads either.

## Suggested fix

If the intended semantics are "the health held when the duel was accepted", snapshot both alongside the positions that are already snapshotted:

```kotlin
val senderHealth = sender.health
val targetHealth = target.health
```

A maintainer decision on the intended semantics is wanted before that is applied.

## Provenance

This was noticed while reviewing PR #2019.

---

This issue was filed during a Gardener session (https://github.com/Stephenson-Software/gardener).

Issue #2020, opened 2026-08-19
MfFactionMapCommand reads sender.world off the main thread inside its async task


dmccoystephenson:
## Summary

`MfFactionMapCommand` reads `sender.world` from inside the `Runnable` handed to `runTaskAsynchronously`, at `src/main/kotlin/com/dansplugins/factionsystem/command/faction/map/MfFactionMapCommand.kt:65`:

```kotlin
val map = renderMap(faction, mapType, sender.world, senderChunkX - 10, senderChunkZ - 4, senderChunkX + 10, senderChunkZ + 4)
```

The chunk coordinates immediately above it are handled correctly — `sender.location.chunk` is read at line 46, on the main thread, before dispatching, and only the resulting `x`/`z` ints cross into the task. The world reference was not given the same treatment.

## Why it matters

Two consequences follow, mirroring the ones described in #2006:

- **The world is sampled at task time, not at command time.** A player who changes world between issuing `/faction map` and the task running is rendered a map of claims in the world they arrived in, indexed by chunk coordinates taken from the world they left. The rendered grid is then a mix of two worlds and shows claims that are not where the player is standing.
- **The read itself is off-main-thread Bukkit API access.** `Player.getWorld()` is not documented as thread-safe, and the returned `World` is passed straight into `claimService.getClaim(world, x, z)` for every cell of a 21x9 grid.

## Suggested fix

Snapshot the world alongside the chunk coordinates that are already snapshotted, and have the task use the snapshotted reference:

```kotlin
val senderChunk = sender.location.chunk
val senderWorld = senderChunk.world
val senderChunkX = senderChunk.x
val senderChunkZ = senderChunk.z
```

That keeps the world and the coordinates consistent with one another, since both are then taken from the same `Location` read.

## Provenance

This was noticed while surveying the command package for PR #2019, which closes #2006. It is not covered by that issue's inventory, so it was filed separately rather than folded into that changeset.

---

This issue was filed during a Gardener session (https://github.com/Stephenson-Software/gardener).

Issue #2018, opened 2026-08-17
DEPLOY_BRANCH is pinned to a deleted copilot branch, so no push to main has ever deployed


dmccoystephenson:
## Summary

The repository variable `DEPLOY_BRANCH` is set to `copilot/add-github-actions-workflow`, a branch that no longer exists. `.github/workflows/deploy.yml` gates its job on:

```
if: github.ref_name == (vars.DEPLOY_BRANCH || 'main') || workflow_dispatch
```

so every push to `main` evaluates that condition as false and the deploy job is skipped. The last five recorded runs of `deploy.yml` — including two on `main` — all show `skipped`.

## Impact

No merge to `main` has ever deployed to the omcsi instance through this workflow. The workflow reports as present and green, which is why the gap is invisible from the Actions tab: a skipped job is not a failed one.

## Why this is being filed rather than fixed

Correcting the variable to `main` would mean the *next* merge auto-deploys to the live server, carrying every undeployed change since the variable was set. That is a deployment decision, not a repository-hygiene one, and it belongs to whoever operates that server. The same defect was found and corrected in `Dans-Plugins/Activity-Tracker` (its issue #95), where flipping the variable did exactly that — so the consequence here is known rather than speculative.

## Suggested resolution

Either:

1. Set `DEPLOY_BRANCH` to `main` **after** a controlled `workflow_dispatch` run has confirmed the deploy path still works against the current server, or
2. Remove the variable entirely so the workflow falls back to its documented `'main'` default, again after a manual dispatch.

Whichever is chosen, a follow-up worth considering is failing loudly when `DEPLOY_BRANCH` names a branch that does not exist, so a stale pin cannot masquerade as a healthy skip.

Found while verifying that merging #2016 into `main` could not trigger an unintended production deploy.

---

_drafted by Claude on behalf of Daniel Stephenson_

Issue #2015, opened 2026-08-15
CHANGELOG.md has two separate '### Fixed' sections under '## [Unreleased]'


dmccoystephenson:
## Summary

`CHANGELOG.md` currently carries two separate `### Fixed` sections under the same `## [Unreleased]` heading — one immediately after `## [Unreleased]` and before `### Added`, and a second after `### Changed`.

## Current shape

```
## [Unreleased]

### Fixed
- The `Dev Release` workflow now retries publishing ...

### Added
- A JSON file storage backend ...
- ...

### Changed
- `/faction addmember` no longer silently refuses ...

### Fixed
- `/faction unclaim`, `/faction gate remove` and `/faction claim check` ...
- ...
```

## Why it matters

Keep a Changelog expects one section per change type per release. A duplicated heading means a reader scanning for fixes finds the first block and stops, and it makes the next release's notes wrong if they are assembled by section. It also gives every future pull request two plausible places to add a `Fixed` entry, so the split tends to persist and widen.

## Suggested fix

Merge the two into a single `### Fixed` block in the conventional Keep a Changelog order (`Added`, `Changed`, `Deprecated`, `Removed`, `Fixed`, `Security`). This is a pure move of existing lines with no wording changes, and is best done on its own rather than folded into a feature branch, where the moved paragraphs would bury the branch's real change.

This was noticed while grooming #2001 and deliberately left alone there for that reason.

---

_drafted by Claude on behalf of Daniel Stephenson_

PR #2023, opened 2026-08-27
Restore main's test compilation by passing EntityInteractionProtection in PlayerQuitListenerTest

### LINKED ISSUE #2022: main fails to compile its tests: PlayerQuitListenerTest was not updated for the constructor change in #2001
## Summary

The `Build` check on `main` has been failing since b3e457d4 (2026-08-22), the merge of #2001. Test compilation fails, so no `./gradlew test` run — locally or in CI — has succeeded on `main` for five days, and every pull request opened against `main` inherits a red required check that has nothing to do with its own changes.

The failure is:

```
e: src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt: (58, 40): No value passed for parameter 'entityInteractionProtection'
```

Reproduced locally at `b3e457d4` with `./gradlew compileTestKotlin`, and present in CI [run 32547876409](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/32547876409).

## Cause

This is a semantic merge conflict between two changes that do not overlap in any file, so neither the merge nor either branch's own CI could have caught it:

- #2016 (merged as a50d558f, 2026-08-16) added `PlayerQuitListenerTest.kt`, which constructs the listener as `PlayerQuitListener(plugin)`.
- #2001 (merged as b3e457d4, 2026-08-22) added a second constructor parameter, `entityInteractionProtection: EntityInteractionProtection`, to `PlayerQuitListener`. Its branch was cut before #2016 landed, so the new test file was not present on it and was never updated.

`PlayerQuitListenerTest.kt` is touched by neither side of the merge, which is why `git` reported no conflict.

## Scope of the breakage

Only test sources are affected. The main source set still compiles, which is why the `Dev Release` workflow reported success on the same commit and the published `dev` prerelease JAR is not itself broken. The consequence is a blind spot rather than a shipped defect: the whole suite has been unrunnable, so no regression in any of the 580-plus tests would have been noticed during that window.

## Suggested fix

Pass a mocked `EntityInteractionProtection` to the constructor in the test's `setUp`. The `forgetPlayer` call that #2001 added to the listener also has no test covering it, and is worth asserting while the mock is being wired in.

---

---

dmccoystephenson:
Closes #2022

## What was wrong

`main` has not compiled its test sources since b3e457d4 (2026-08-22), the merge of #2001. The `Build` check has been red on `main` for five days, and every pull request opened against `main` since then inherits that failure as a required check unrelated to its own changes.

```
e: src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt: (58, 40): No value passed for parameter 'entityInteractionProtection'
```

It is a semantic merge conflict between two changes with no file in common, which is why no conflict was reported and neither branch's own CI could have caught it:

- #2016 (a50d558f, 2026-08-16) added `PlayerQuitListenerTest.kt`, constructing the listener as `PlayerQuitListener(plugin)`.
- #2001 (b3e457d4, 2026-08-22) added a second constructor parameter, `entityInteractionProtection`. That branch was cut before the test file existed, so the call site was never updated.

Only test sources are affected — the main source set still compiles, which is why `Dev Release` reported success on the same commit and the published `dev` JAR is not itself broken. The cost is a blind spot rather than a shipped defect: the suite has been unrunnable, so a regression in any of the 580-plus tests would have gone unnoticed during that window.

## What was changed

- A mocked `EntityInteractionProtection` is created in `setUp` and passed to the listener under test. `mockito-inline` is already on the test classpath, so the final Kotlin class is mockable without production changes.
- One test is added, asserting that `forgetPlayer` is called with the quitting player's UUID before the listener returns. That call was added by #2001 and had no coverage; wiring the collaborator in as a mock is what makes asserting it cheap, so it is done here rather than deferred.

No production code is touched by this changeset. No user-facing behaviour changes, so no `CHANGELOG.md` entry is added.

## Data safety

No schema migration, no DPC contract change, no `config.yml` default change, no `plugin.yml` change, no `lang/` change. The diff is confined to one test file.

## What was deliberately left out

The rest of the open backlog is deferred, since one unit of work is taken per cycle and restoring a red required check on `main` was ranked ahead of everything else available. Specifically:

- #2020 (`MfFactionMapCommand` reads `sender.world` off the main thread) — ready to implement and appropriately scoped, but folding an unrelated bug fix into a build-unblocking change would delay the latter behind the review of the former.
- #2015 (duplicate `### Fixed` sections in `CHANGELOG.md`) — that issue explicitly asks for the move to be made on its own branch, since the relocated paragraphs would bury whatever else the branch contains. Honoured here rather than folded in.
- #2021 (`MfDuelAcceptCommand` reads both players' health off the main thread) — blocked on a maintainer decision about which health a duel is meant to restore, as recorded on the issue.
- #2018 (`DEPLOY_BRANCH` pinned to a deleted branch) — modifies `.github/workflows`, which is on this loop's do-not-auto-merge list and wants a human.

## Test plan

- [x] `./gradlew compileTestKotlin` — fails on `main` at b3e457d4, passes on this branch.
- [x] `./gradlew clean test` — `BUILD SUCCESSFUL`, whole suite green. This is the first passing full run since 2026-08-22.
- [x] `./gradlew lintKotlin` — `BUILD SUCCESSFUL`.
- [x] The new `forgetPlayer` assertion was empirically confirmed to fail without the production behaviour: removing the `entityInteractionProtection.forgetPlayer(...)` line from `PlayerQuitListener` locally produced `8 tests completed, 1 failed` with `WantedButNotInvoked` at `PlayerQuitListenerTest.kt:176`, and restoring it returned the class to green. The production file is unmodified in this changeset.

---

### DIFF
--- a/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt
+++ b/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt
@@ -164,8 +164,10 @@ class PlayerQuitListenerTest {
 
     @Test
     fun onPlayerQuit_ShouldDropRetainedEntityNotificationStateOnTheServerThread() {
-        // Arrange
+        // Arrange - the UUID is read up front rather than inside the verify() argument list, so that
+        // no invocation on another mock is made while verification mode is armed
         `when`(playerService.getPlayer(fixture.player)).thenReturn(fixture.mfPlayer)
+        val playerUuid = fixture.player.uniqueId
 
         // Act
         uut.onPlayerQuit(fixture.event)
@@ -173,7 +175,7 @@ class PlayerQuitListenerTest {
         // Assert - the retained notification is keyed on the player's UUID, so it has to be dropped
         // before the listener returns, or a reconnecting player inherits the de-duplication window
         // that was open when they left
-        verify(entityInteractionProtection).forgetPlayer(fixture.player.uniqueId)
+        verify(entityInteractionProtection).forgetPlayer(playerUuid)
     }
 
     @Test
--- a/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt
+++ b/src/test/kotlin/com/dansplugins/factionsystem/listener/PlayerQuitListenerTest.kt
@@ -38,6 +38,7 @@ class PlayerQuitListenerTest {
     private lateinit var plugin: MedievalFactions
     private lateinit var playerService: MfPlayerService
     private lateinit var interactionService: MfInteractionService
+    private lateinit var entityInteractionProtection: EntityInteractionProtection
     private lateinit var teleportService: MfTeleportService
     private lateinit var logger: Logger
     private lateinit var config: FileConfiguration
@@ -55,7 +56,8 @@ class PlayerQuitListenerTest {
         mockLogger()
         mockScheduler()
         `when`(playerService.save(anyMfPlayer())).thenReturn(Success(fixture.mfPlayer))
-        uut = PlayerQuitListener(plugin)
+        entityInteractionProtection = mock(EntityInteractionProtection::class.java)
+        uut = PlayerQuitListener(plugin, entityInteractionProtection)
     }
 
     @Test
@@ -158,6 +160,20 @@ class PlayerQuitListenerTest {
         assertEquals(fixture.playerId, saved.id)
         assertEquals(20.0, saved.power)
         assertEquals(20.0, saved.powerAtLogout)
+    }
+
+    @Test
+    fun onPlayerQuit_ShouldDropRetainedEntityNotificationStateOnTheServerThread() {
+        // Arrange
+        `when`(playerService.getPlayer(fixture.player)).thenReturn(fixture.mfPlayer)
+
+        // Act
+        uut.onPlayerQuit(fixture.event)
+
+        // Assert - the retained notification is keyed on the player's UUID, so it has to be dropped
+        // before the listener returns, or a reconnecting player inherits the de-duplication window
+        // that was open when they left
+        verify(entityInteractionProtection).forgetPlayer(fixture.player.uniqueId)
     }
 
     @Test


dmccoystephenson:
Self-review rubric, scored against the diff and against command output rather than against judgement:

- **Scope: PASS** — one file is modified, `PlayerQuitListenerTest.kt`, which is exactly the file named by #2022. No production file is touched; `git diff --name-only origin/main...HEAD` returns that single path.
- **Tests-new: PASS** — no new production method is introduced, so nothing is owed here. The one test added covers the `forgetPlayer` call that #2001 introduced and left uncovered.
- **Tests-fix: PASS** — confirmed empirically rather than by reasoning. Removing the `entityInteractionProtection.forgetPlayer(...)` line from `PlayerQuitListener` locally produced `8 tests completed, 1 failed`, `WantedButNotInvoked` at `PlayerQuitListenerTest.kt:176`; restoring it returned the class to green. The production file is unmodified in the pushed changeset. The compile fix itself is verified the same way: `./gradlew compileTestKotlin` fails at `main`'s `b3e457d4` and passes here.
- **Sibling structure: PASS, after a correction.** The first push asserted `verify(entityInteractionProtection).forgetPlayer(fixture.player.uniqueId)`, invoking a mock inside a `verify` argument list. No other assertion in this class does that, and it relies on Mockito declining to consume the armed verification mode for an invocation on a different mock — true today, but not a property this test should depend on. The read is now hoisted to a local before the act step, in `556c4c05`.
- **Sibling renames: PASS** — no identifier is renamed.
- **Docs: PASS** — no row of the documentation sources-of-truth table applies. The change is test-only and alters no user-facing behaviour, so no `CHANGELOG.md` entry is warranted; no command, config option, faction flag, permission node, dependency or schema is touched.
- **Migrations-additive: PASS** — `git diff origin/main...HEAD -- '*db/migration/*'` is empty.
- **Lang-externalized: PASS** — no string literal is added to any committed `.kt` or `.java` production file.
- **DPC-contract: PASS** — nothing under `dpc/` or `pacts/` is in the diff.
- **Main-thread-safety: PASS** — no Bukkit API access is added. The new assertion exercises the existing server-thread path, and is placed before the queued async tasks are drained, matching how the sibling `onPlayerQuit_ShouldUnloadInteractionStatusOnTheServerThread` establishes the same property.
- **Plugin.yml: PASS** — no command or permission is added.
- **Config-defaults: PASS** — `config.yml` is not in the diff.
- **Issue resolution: PASS** — the surface area #2022 names, the one-argument constructor call in `setUp`, is the line changed. Nothing in that issue is left partially addressed.
- **CI: PASS** — `build`, `changes` and `docker-build` all passed on `120b4619` ([run 33054433577](https://github.com/Dans-Plugins/Medieval-Factions/actions/runs/33054433577)), which is itself the proof of the fix, since `build` is failing on `main`. Re-confirmation on `556c4c05` is pending at the time of writing and is treated as the merge gate.

One observation is folded in here rather than left inline, since it concerns no line this changeset adds. The failure mode that caused this is not addressable in code: two branches with no file in common diverged, `git` reported no conflict, and each branch's own CI was green before merge. Requiring branches to be up to date with `main` before merging, or making `build` a required check on the merge queue rather than only on the head, would have caught it. That is a repository-settings decision for the maintainer, and has been recorded as a self-audit finding rather than attempted here.

---

---

PR #2025, opened 2026-08-29
fix: snapshot the sender's world on the main thread in /faction map

### LINKED ISSUE #2015: CHANGELOG.md has two separate '### Fixed' sections under '## [Unreleased]'

### LINKED ISSUE #2020: MfFactionMapCommand reads sender.world off the main thread inside its async task

dmccoystephenson:
## Summary

- **`/faction map` now renders the world the sender stood in when the command was run.** The chunk coordinates the grid is centred on were already read from `sender.location` on the main thread before the lookup was queued, but the world was read as `sender.world` from inside the queued `Runnable`. A player who changed world in between was therefore shown a grid of the world they arrived in, indexed by chunk coordinates taken from the world they left — a mix of two worlds showing claims that are nowhere near where they are standing. The world is now taken from the same `Location` read that already supplied the coordinates (`senderChunk.world`), so the two are always consistent with one another.
- **The read was also off-main-thread Bukkit API access.** `Player.getWorld()` is not documented as thread-safe, and the returned reference was passed straight into `claimService.getClaim(world, x, z)` for every one of the grid's 189 cells. This is the same class of defect that #2006 covered for positions and that PR #2019 finished; the world reference had not been given the same treatment.
- **The two `### Fixed` sections that had accumulated under `## [Unreleased]` in `CHANGELOG.md` are merged into one.** Keep a Changelog expects one section per change type per release, and the split gave every future PR two plausible places to add a `Fixed` entry. The existing entry has been moved, not reworded.

## Changes

| File | Change |
| --- | --- |
| `command/faction/map/MfFactionMapCommand.kt` | The world is snapshotted alongside the chunk coordinates and the snapshot is passed to `renderMap` (net +1 line). |
| `command/faction/map/MfFactionMapCommandTest.kt` | New test class, mirroring `MfFactionClaimCircleCommandTest`. |
| `CHANGELOG.md` | The duplicate `### Fixed` heading is removed and its entry moved into the surviving section; an entry for this fix is added. |

## Test plan

- [x] `./gradlew clean test` — BUILD SUCCESSFUL, full suite.
- [x] `./gradlew lintKotlin` — BUILD SUCCESSFUL.
- [x] Regression evidence obtained empirically rather than by reasoning: with the one-line source fix stashed (`git stash push -- <the command file>`), `testOnCommand_rendersTheWorldTheSenderStoodInWhenTheCommandWasRun` FAILS with `ArgumentsAreDifferent`, and the other three tests in the class still pass. With the fix restored, all four pass. The two worlds are distinct mocks and the sender is moved between the command call and the task drain, so the queued task can observe either one.
- [x] Coverage added for the permission refusal and non-player-sender paths, and for the grid being centred on the sender's chunk (`-10,-4` to `+10,+4`), neither of which was previously tested — `MfFactionMapCommand` had no test class at all.

## Data safety

No schema migrations; no DPC contract change; no `config.yml` or `plugin.yml` change; no `lang/` change (no user-facing string is added or altered). No path on the do-not-auto-merge list is touched. Which claims a stationary player is shown is unchanged.

## Deferred this cycle

- **#2021** (`MfDuelAcceptCommand` reads both players' health off the main thread) — deferred deliberately, not overlooked. The issue itself states that a maintainer decision on the intended semantics is wanted first: moving the health read earlier changes *which* health a duel restores, which is a design decision rather than a timing correction.
- **#1918** and **#1915** (the two open `copilot/*` drafts) — both are conflicting against `main` and large (698/483 and 2083 changed lines respectively), so neither fits inside this cycle's grooming budget. #1918 additionally changes config defaults, which is charter-gated for a human reviewer.
- The remainder of the open backlog was not surveyed for this cycle, which was scoped to the two issues above.

Closes #2020
Closes #2015

---


dmccoystephenson:
Self-review rubric, anchored on CI green at head `48d97759` (`build` pass 3m26s, `docker-build` pass 5m32s, `changes` pass).

- **Scope: PASS** — `git diff --name-only origin/main...HEAD` lists exactly three files, and each is required: the one-line source fix, its new test class, and the changelog. No formatting, rename or comment churn is present.
- **Tests-new: PASS** — no new public method is introduced; the changed method (`onCommand`) had no test class at all before this PR and now has four, covering the permission refusal, the non-player sender, the grid extents, and the world snapshot.
- **Tests-fix: PASS** — established empirically, not by reasoning. With the source fix stashed, `testOnCommand_rendersTheWorldTheSenderStoodInWhenTheCommandWasRun` fails with `ArgumentsAreDifferent` at `MfFactionMapCommandTest.kt:133` while the other three tests still pass; with the fix restored, all four pass.
- **Sibling structure: PASS** — the new test class mirrors `MfFactionClaimCircleCommandTest`: `@TestInstance(PER_CLASS)`, `TestUtils` fixtures, the same prepare/execute/verify comment layout, and the same queued-task scheduler mock and `runPendingTasks` drain.
- **Sibling renames: no signal** — nothing is renamed by this PR.
- **Docs: PASS** — a `Fixed` entry is added under `## [Unreleased]`. `COMMANDS.md:176-183` was re-read against the implementation and remains accurate: the permission (`mf.map`), the aliases and all three usage forms are unchanged by this PR. `CONFIG.md`, `FACTION_FLAGS.md`, `DATABASE_QUERYING.md`, `DEPENDENCY_USAGE.md`, `USER_GUIDE.md` and both READMEs describe nothing this PR alters.
- **Issue resolution: PASS** — #2020 names `MfFactionMapCommand.kt:65` and the `sender.world` read specifically, and that read is the line changed. #2015 names the duplicated `### Fixed` heading, and the diff removes one heading and moves its single entry into the survivor with no rewording.
- **Migrations-additive: PASS** — `git diff origin/main...HEAD -- '*db/migration/*'` returns empty output; no migration file is touched.
- **Lang-externalized: PASS** — the source diff is two lines, neither of which contains a string literal.
- **DPC-contract: PASS** — no path under `dpc/` and no pact file appears in the diff.
- **Main-thread-safety: PASS, with a residual noted below** — the read named by #2020 is moved ahead of the dispatch and no new off-thread Bukkit access is introduced. It should not be inferred from this that the command is now free of off-thread access; see the first finding.
- **Plugin.yml: no signal** — no command or permission is added or changed.
- **Config-defaults: PASS** — `src/main/resources/config.yml` is not in the diff.

## Findings

`src/main/kotlin/com/dansplugins/factionsystem/command/faction/map/MfFactionMapCommand.kt:106,121,126,132` — the async task still reads `plugin.config.getString("wilderness.color")` and `plugin.flags.color` from inside `renderMap`/`getColor`, once per grid cell, so up to 189 configuration reads happen off the server thread per invocation. This predates the PR and is repo-wide rather than specific to this command, so it is deliberately left out of a two-line fix, but it means the changelog entry should be read as "the world reference is now consistent", not as "this command no longer touches anything off-thread". No issue has been filed for it, since the same pattern would have to be addressed across the command package rather than here.

`src/test/kotlin/com/dansplugins/factionsystem/command/faction/map/MfFactionMapCommandTest.kt` — coverage is confined to the `NORMAL` map type. The `DIPLOMATIC` branch of `getColor`, which is where the relationship logic lives (at-war, ally, vassal tree, liege chain), remains untested, as does the `CommandFactionMapMapTypeRequiresFaction` refusal that a factionless sender receives for that type. Those paths are untouched by this fix, so testing them is not required to close #2020, but the class is not a complete characterization of the command and should not be treated as one.

`src/test/kotlin/com/dansplugins/factionsystem/command/faction/map/MfFactionMapCommandTest.kt:150` — `player.world` is stubbed by `stubSenderChunk` even though the fixed code no longer reads it. This is intentional and load-bearing rather than a leftover: the stub is what lets the regression test observe the pre-fix behaviour, since the unfixed code reads `player.world` at task time and would otherwise throw rather than return the wrong world. It is called out here so it is not later removed as dead stubbing.

This review was drafted during a Gardener session (https://github.com/Stephenson-Software/gardener).

---

Issue #2037, opened 2026-09-19
PostgreSQL is documented but no PostgreSQL driver is shipped; CONFIG.md dialect spellings do not match the code


dmccoystephenson:
Found while running the fleet's save-compatibility gate across database backends (runs [35464461564](https://github.com/Dans-Plugins/release-gates/actions/runs/35464461564) v5.8.1 and [35464665095](https://github.com/Dans-Plugins/release-gates/actions/runs/35464665095) 6.0.0 candidate).

1. **No PostgreSQL driver in either jar.** `build.gradle` shades `mariadb-java-client` and `h2` only, while CONFIG.md and MIGRATION_GUIDE.md list PostgreSQL and the 6.0.0 code references `org.postgresql.Driver`. With `jdbc:postgresql://…` the plugin fails at enable: `java.lang.ClassNotFoundException: org.postgresql.Driver` (6.0.0) / `No suitable driver` (5.8.1). The Flyway SQL is also MySQL-flavoured (backticks, `json`), so a driver alone would not make it work. Either ship and test the backend, or remove it from the documentation so an operator does not configure it and lose their first startup.
2. **`database.dialect` values in CONFIG.md are wrong.** The code resolves the dialect with `SQLDialect.valueOf`, so the accepted spellings are `H2`, `MYSQL`, `MARIADB`, `POSTGRES`; CONFIG.md documents `MySQL` / `PostgreSQL`, which throw.
3. **v5.8.1 cannot start on MariaDB** (`No suitable driver` for `jdbc:mariadb://` — `onEnable` only calls `Class.forName("org.h2.Driver")`, the relocated MariaDB driver is never registered; PR #1759 addressed this in 2023 and was closed unmerged). The 6.0.0 candidate loads the driver by URL prefix and runs on MariaDB 11.8 ([35464660832](https://github.com/Dans-Plugins/release-gates/actions/runs/35464660832)). Worth a line in the 6.0.0 release notes: MariaDB becomes usable for the first time.
4. Flyway 9.4 warns on MariaDB 11.8 (`newer than this version of Flyway … latest supported 10.6`); migrations ran, but a Flyway bump would silence it.

---

_drafted by Claude on behalf of Daniel Stephenson_

dmccoystephenson:
Items 1 and 2 are addressed on `main` by #2038 (`7dcae075`):

[...]

Issue #2036, opened 2026-09-19
Release blocked: v6.0.0 failed the save-compatibility gate (stop-1)


dmccoystephenson:
The candidate for `v6.0.0` (`e6c1dc84bbf280bd836bb15c7f4a7041e569322f`) was given the data folder its current stable release wrote on a real Spigot server and did not pass the save-compatibility gate, so no stable release was published.

**Gate:** save-compatibility gate (`save-compat`)

**Failed check:** `stop-1` — database did not close cleanly: trace file medieval_factions_db.trace.db

**Evidence:** https://github.com/Dans-Plugins/release-gates/actions/runs/35461094920 (the run's artifact holds the full server log and the plugin's data folder)

Gates that passed first: t1: [boot run](https://github.com/Dans-Plugins/release-gates/actions/runs/35460950593)

The next attempt is made automatically once `main` moves. See [Release channels](https://github.com/Dans-Plugins/dpc-conventions/blob/main/docs/RELEASE_CHANNELS.md) for what the gates check.

---

_drafted by Claude on behalf of Daniel Stephenson_

dmccoystephenson:
The `medieval_factions_db.trace.db` the gate captured after the candidate's first shutdown (artifact of https://github.com/Dans-Plugins/release-gates/actions/runs/35461094920):

```
com.dansplugins.factionsystem.shadow.org.h2.message.DbException: (Message 90028 not found) [90028-214]
	at com.dansplugins.factionsystem.shadow.org.h2.message.DbException.get(DbException.java:212)
	at com.dansplugins.factionsystem.shadow.org.h2.mvstore.db.Store.close(Store.java:402)
	at com.dansplugins.factionsystem.shadow.org.h2.engine.Database.closeOpenFilesAndUnlock(Database.java:1261)
	at com.dansplugins.factionsystem.shadow.org.h2.engine.OnExitDatabaseCloser.onShutdown(OnExitDatabaseCloser.java:85)
```

[...]

PR #2038, opened 2026-09-20
Open embedded H2 with DB_CLOSE_ON_EXIT=FALSE; accept documented dialect spellings; database docs match the shipped drivers

dmccoystephenson:
## Summary

Two hardening changes to how the plugin opens its database, and the database documentation brought in line with what the jar ships. Refs #2036 (the `DB_CLOSE_ON_EXIT` recommendation) and #2037 (dialect spellings, PostgreSQL).

- **Embedded H2 is opened with `DB_CLOSE_ON_EXIT=FALSE`.** The setting is appended to `database.url` unless the operator already set it; every other URL is left untouched. By default H2 registers a JVM shutdown hook to close any database still open at exit. The plugin closes its pool — and with it the database — in `onDisable`, while its classloader is still loaded, so the hook adds nothing on the normal path. It only ever ran when something else (a leaked connection, another plugin sharing the file) kept the database open past `onDisable`; by then Bukkit had unloaded the plugin, and the close failed inside an unloaded classloader (`Message 90028 not found`), leaving the store unflushed with a `medieval_factions_db.trace.db` beside it. That is the shape recorded on a live server over 225 shutdowns and reproduced on the release gate with Currencies 2.1.0 installed (#2036). With no hook there is nothing left to fail after unload.
- **`database.dialect` is parsed case-insensitively** and accepts the spellings the docs used (`MySQL`, `MariaDB`, `PostgreSQL`) alongside the enum names. Previously only exact upper-case enum names worked, so a config written from `CONFIG.md` failed at startup. An unknown value still fails at startup, now naming the accepted values.
- **Driver preload covers only the two bundled drivers** (H2, MariaDB Connector/J). The `Class.forName` calls for `com.mysql.cj.jdbc.Driver` and `org.postgresql.Driver` could only ever throw `ClassNotFoundException`, since neither is shipped.
- **Docs** (`CONFIG.md`, `DATABASE_QUERYING.md`, `docs/MIGRATION_GUIDE.md`, `README.md`): PostgreSQL is no longer listed; MariaDB and MySQL servers are both reached through `jdbc:mariadb://` (the bundled driver only takes `jdbc:mysql://` with `?permitMysqlScheme`); the dialect values are the real ones; the `DB_CLOSE_ON_EXIT` behaviour is described.

All three places a datasource is opened (startup, and both legs of `/faction migrate`) go through the new `MfJdbc` helper. No migration, `config.yml` default, `plugin.yml` or DPC code is touched; an existing server's `database.url` is not rewritten — the setting is applied in memory at open.

## Verification

- `./gradlew test lintKotlin` (JDK 17): 634 tests pass, lint clean. `MfJdbcTest` covers the URL hardening (default URL, case, operator override, non-H2 URLs), the dialect parsing (enum names, documented spellings, blank, unknown), and opens an in-memory H2 with the hardened URL plus the default's `MODE=MYSQL;DATABASE_TO_UPPER=false` to prove the bundled H2 2.1.214 accepts the combination.
- Not exercised here: a real-server stop/start cycle over an existing `medieval_factions_db.mv.db`. The release gates run that (save-compatibility over the real-world and bot-recorded fixtures with five restart cycles, H2 and JSON backends) against the `dev` build once this is on `main`, before any stable release is considered.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016xxuhDAUQuygtx17Gi7DeM


dmccoystephenson:
**Self-review (pre-merge)**

- **Scope / data safety:** no migration, no `config.yml` default, no `plugin.yml`, no DPC code touched; `database.url` on disk is never rewritten — the setting is applied to the in-memory URL at open. An operator who already set `DB_CLOSE_ON_EXIT` (either value) is left alone; MariaDB URLs are untouched.
- **Behaviour on the normal path is unchanged:** `onDisable` already closed the pool, which closes the database; the only difference is that no H2 shutdown hook exists afterwards. The hook was only ever reached when the database was still open after `onDisable`, and in that situation it failed (#2036).
- **`/faction migrate`** opens two datasources through the same helper, so a migration's H2 side is hardened identically; the dialect is resolved the same way at all three sites.
- **Risk accepted:** `Class.forName` for the MySQL and PostgreSQL driver names is gone. Both could only throw `ClassNotFoundException` (neither driver is bundled); an unsupported URL now fails from the connection pool with the driver-lookup error instead of from the preload. Net effect for an operator is the same failure with a clearer message.
- **Verified:** `./gradlew test lintKotlin` on JDK 17 — 634 tests, lint clean; CI `build` and `docker-build` green on the head. A real stop/start over an existing H2 store is left to the release gates on the `dev` build (five restart cycles over the real-world and bot fixtures), which run before any stable is considered.

---

PR #2039, opened 2026-09-20
Leave AUTO_SERVER H2 URLs alone: H2 refuses them with DB_CLOSE_ON_EXIT=FALSE

dmccoystephenson:
## What went wrong

#2038 appended `;DB_CLOSE_ON_EXIT=FALSE` to every H2 URL. H2 2.1.214 refuses that together with `AUTO_SERVER=TRUE` — `Feature not supported: "AUTO_SERVER=TRUE && DB_CLOSE_ON_EXIT=FALSE" [50100-214]` (reproduced directly against the bundled H2 jar) — and the default `database.url` is in auto-server mode so that other plugins can share the file. So the `dev` build from `7dcae075` fails at enable on the default config: `HikariPool$PoolInitializationException: Failed to initialize pool: (Message HYC00 not found) [50100-214]`. The save-compatibility gate caught it on the candidate's first boot over the real-world fixture (release-gates run 35481312415), after the baseline, the bot scenario and the fixture checks had all passed. No stable release carries it; the `dev` prerelease does, for about an hour.

The unit test that "proved H2 accepts the URL" opened a `mem:` store, which is never in auto-server mode — that is why it passed.

## Fix

`MfJdbc.hardenUrl` leaves a URL alone when it contains `AUTO_SERVER=TRUE` (case-insensitive, whitespace-tolerant); an operator's explicit `AUTO_SERVER=FALSE` or an embedded URL without the flag still gets the setting. The explicit pool close in `onDisable` — which is what actually keeps the exit hook from having work to do — is unchanged. `CONFIG.md` and the changelog say so.

## Verification

- `MfJdbcTest`: the default URL is returned unchanged; `auto_server = true` and `AUTO_SERVER=FALSE` variants; and a new test opens **file** stores in a temp directory — the default auto-server URL as the helper returns it, and an embedded URL with the setting appended — through the bundled H2, so the exact combination that failed is now exercised.
- `./gradlew test lintKotlin` (JDK 17): 636 tests, lint clean.
- After merge the gate profile is re-run on the new `dev` build.

---

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_016xxuhDAUQuygtx17Gi7DeM


dmccoystephenson:
**Self-review (pre-merge)**

- Scope: one guard in `MfJdbc.hardenUrl`, tests, `CONFIG.md`, changelog. Nothing else from #2038 changes; no config default, migration or `plugin.yml`.
- The default URL now round-trips through the helper byte-for-byte, so a server on the default config opens its database exactly as 5.8.1 and the pre-#2038 dev builds did.
- The new test opens file stores (auto-server and embedded) through the bundled H2 — the exact shape that failed — rather than `mem:`, which is never auto-server.
- Verified: `./gradlew test lintKotlin` (JDK 17): 636 tests, lint clean; CI `build` + `docker-build` green on the head. The save-compatibility gate is re-run on the resulting `dev` build straight after merge.

---

